PR #5, opened 2017-06-20
Fix all eslint issues

c-w:
This pull request fixes all eslint issues and sets up CI so that we always run tests and lint on every check-in/pull-request.

The pull request also fixes a few bugs uncovered by eslint, mostly to do with referencing undefined variables.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,15 @@
+language:
+  node_js
+
+node_js:
+  - "7"
+
+install:
+  - npm install
+
+env:
+  - NPM_COMMAND="lint"
+  - NPM_COMMAND="test"
+
+script:
+  - npm run $NPM_COMMAND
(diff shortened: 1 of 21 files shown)


.travis.yml
```diff
@@ -0,0 +1,15 @@
+language:
+  node_js
+
+node_js:
+  - "7"
```
Smarker, line 5: package.json has node version 6.0.0 not sure if we want to keep our version at 7 or change what is in .travis.yml from 7 to 6.

Smarker (approved):
Looks good. I see that I will have to make changes in my code since I might have used double quotes a couple of times rather than single quotes.

.travis.yml
c-w: Good catch. We can run on both actually, see 0aa6524.

c-w: NB: you can use `./node_modules/.bin/eslint src --fix` to auto-fix simple issues like double-quotes vs single-quotes.

erikschlegel (commented):
Thank you!

PR #6, opened 2017-06-22
Crisis lex integration

Smarker:
This pull requests allows [crisis lex topics](https://github.com/sajao/CrisisLex/tree/master/data) to be pulled based on siteType from azure blob storage into Cassandra using [datastax nodejs-driver](https://github.com/datastax/nodejs-driver).

See **/resolvers/Settings.js: createOrReplaceSite** for most of my changes.

I created a temporary TopicsSchema so that I could make graphql /get and /list queries for topics. If there is a change to the TopicsSchema, we will need to update **CassandraTableStorageManager.prepareInsertTopic** to account for the schema change.

One more major change was the addition of:
const Promise = require('bluebird');
const azure = Promise.promisifyAll(require('azure-storage'));
I felt that changing it from cb style to promise style made the code a little easier to read.

Another minor addition was mocha unit tests. They definitely can be improved with mocks, stubs, etc.

Another note: I was able to insert topics into my local Cassandra cluster.

I added some environment variables to my FortisEnv.sh, so you might want to as well:
**BLOB_STORAGE_CONNECTION_STRING
CASSANDRA_KEYSPACE
CASSANDRA_CONTACT_POINTS**

**Pending:** 
- add appinsights
- use sinon in unit tests
- finalizing options for Cassandra client

**Questions**
- createOrReplaceSite takes two arguments (args, res). What is *res*?

### DIFF
--- a/server.js
+++ b/server.js
@@ -22,17 +22,17 @@ var TopicsResolver = require('./src/resolvers/Topics');
 var app = express();
 
 app.use(function(req, res, next) {
-    res.header('Access-Control-Allow-Origin', '*');
-    res.header('Access-Control-Allow-Credentials', true);
-    res.header('Access-Control-Allow-Methods', 'OPTIONS,GET,POST,PUT,DELETE');
-    res.header('Access-Control-Allow-Headers', 'Content-Type, Authorization, x-ms-meta-data*,x-ms-meta-target*,x-ms-meta-abc');
+  res.header('Access-Control-Allow-Origin', '*');
+  res.header('Access-Control-Allow-Credentials', true);
+  res.header('Access-Control-Allow-Methods', 'OPTIONS,GET,POST,PUT,DELETE');
+  res.header('Access-Control-Allow-Headers', 'Content-Type, Authorization, x-ms-meta-data*,x-ms-meta-target*,x-ms-meta-abc');
 
    // intercept OPTIONS method
-    if ('OPTIONS' == req.method) {
-        res.sendStatus(200);
-    } else {
-        next();
-    }
+  if ('OPTIONS' == req.method) {
+    res.sendStatus(200);
+  } else {
+    next();
+  }
 });
 
 console.log('PORT: ' + port);
@@ -41,39 +41,39 @@ app.use(bodyParser.json({limit: '50mb'}));
 app.use(bodyParser.urlencoded({limit: '50mb', extended: true}));
 
 app.use('/api/messages', graphqlHTTP({
-    schema: SpatialMessageSchema,
-    rootValue: SpatialMessageResolver,
-    graphiql: true
+  schema: SpatialMessageSchema,
+  rootValue: SpatialMessageResolver,
+  graphiql: true
 }));
 
 app.use('/api/facts', graphqlHTTP({
-    schema: FactsSchema,
-    rootValue: FactsResolver,
-    graphiql: true
+  schema: FactsSchema,
+  rootValue: FactsResolver,
+  graphiql: true
 }));
 
 app.use('/api/edges', graphqlHTTP({
-    schema: EdgesSchema,
-    rootValue: EdgesResolver,
-    graphiql: true
+  schema: EdgesSchema,
+  rootValue: EdgesResolver,
+  graphiql: true
 }));
 
 app.use('/api/tiles', graphqlHTTP({
-    schema: TileSchema,
-    rootValue: TileResolver,
-    graphiql: true
+  schema: TileSchema,
+  rootValue: TileResolver,
+  graphiql: true
 }));
 
 app.use('/api/settings', graphqlHTTP({
-    schema: SettingsSchema,
-    rootValue: SettingsResolver,
-    graphiql: true
+  schema: SettingsSchema,
+  rootValue: SettingsResolver,
+  graphiql: true
 }));
 
 app.use('/api/topics', graphqlHTTP({
-    schema: TopicsSchema, //TODO change to TopicsSchema
-    rootValue: TopicsResolver,
-    graphiql: true
+  schema: TopicsSchema, //TODO change to TopicsSchema
+  rootValue: TopicsResolver,
+  graphiql: true
 }));
 
 var server = http.createServer(app);
--- a/server.js
+++ b/server.js
@@ -6,16 +6,18 @@ var http = require('http');
 var express = require('express');
 var bodyParser = require('body-parser');
 var graphqlHTTP = require('express-graphql');
-var SpatialMessageSchema = require('./src/schemas/MessageSchema');
-var SpatialMessageResolver = require('./src/resolvers/Messages');
-var FactsSchema = require('./src/schemas/FactSchema');
-var FactsResolver = require('./src/resolvers/Facts');
-var TileSchema = require('./src/schemas/TilesSchema');
-var TileResolver = require('./src/resolvers/Tiles');
-var EdgesSchema = require('./src/schemas/EdgesSchema');
-var EdgesResolver = require('./src/resolvers/Edges');
-var SettingsResolver = require('./src/resolvers/Settings');
-var SettingsSchema = require('./src/schemas/SettingsSchema');
+var SpatialMessageSchema = require("./src/schemas/MessageSchema");
+var SpatialMessageResolver = require("./src/resolvers/Messages");
+var FactsSchema = require("./src/schemas/FactSchema");
+var FactsResolver = require("./src/resolvers/Facts");
+var TileSchema = require("./src/schemas/TilesSchema");
+var TileResolver = require("./src/resolvers/Tiles");
+var EdgesSchema = require("./src/schemas/EdgesSchema");
+var EdgesResolver = require("./src/resolvers/Edges");
+var SettingsResolver = require("./src/resolvers/Settings");
+var SettingsSchema = require("./src/schemas/SettingsSchema");
+var TopicsSchema = require("./src/schemas/TopicsSchema");
+var TopicsResolver = require("./src/resolvers/Topics");
 
 var app = express();
 
@@ -66,6 +68,12 @@ app.use('/api/settings', graphqlHTTP({
     schema: SettingsSchema,
     rootValue: SettingsResolver,
     graphiql: true
+}));
+
+app.use('/api/topics', graphqlHTTP({
+  schema: TopicsSchema, //TODO change to TopicsSchema
+  rootValue: TopicsResolver,
+  graphiql: true,
 }));
 
 var server = http.createServer(app);
--- a/server.js
+++ b/server.js
@@ -71,7 +71,7 @@ app.use('/api/settings', graphqlHTTP({
 }));
 
 app.use('/api/topics', graphqlHTTP({
-  schema: TopicsSchema, //TODO change to TopicsSchema
+  schema: TopicsSchema,
   rootValue: TopicsResolver,
   graphiql: true
 }));
--- a/server.js
+++ b/server.js
@@ -6,18 +6,18 @@ var http = require('http');
 var express = require('express');
 var bodyParser = require('body-parser');
 var graphqlHTTP = require('express-graphql');
-var SpatialMessageSchema = require("./src/schemas/MessageSchema");
-var SpatialMessageResolver = require("./src/resolvers/Messages");
-var FactsSchema = require("./src/schemas/FactSchema");
-var FactsResolver = require("./src/resolvers/Facts");
-var TileSchema = require("./src/schemas/TilesSchema");
-var TileResolver = require("./src/resolvers/Tiles");
-var EdgesSchema = require("./src/schemas/EdgesSchema");
-var EdgesResolver = require("./src/resolvers/Edges");
-var SettingsResolver = require("./src/resolvers/Settings");
-var SettingsSchema = require("./src/schemas/SettingsSchema");
-var TopicsSchema = require("./src/schemas/TopicsSchema");
-var TopicsResolver = require("./src/resolvers/Topics");
+var SpatialMessageSchema = require('./src/schemas/MessageSchema');
+var SpatialMessageResolver = require('./src/resolvers/Messages');
+var FactsSchema = require('./src/schemas/FactSchema');
+var FactsResolver = require('./src/resolvers/Facts');
+var TileSchema = require('./src/schemas/TilesSchema');
+var TileResolver = require('./src/resolvers/Tiles');
+var EdgesSchema = require('./src/schemas/EdgesSchema');
+var EdgesResolver = require('./src/resolvers/Edges');
+var SettingsResolver = require('./src/resolvers/Settings');
+var SettingsSchema = require('./src/schemas/SettingsSchema');
+var TopicsSchema = require('./src/schemas/TopicsSchema');
+var TopicsResolver = require('./src/resolvers/Topics');
 
 var app = express();
 
@@ -71,9 +71,9 @@ app.use('/api/settings', graphqlHTTP({
 }));
 
 app.use('/api/topics', graphqlHTTP({
-  schema: TopicsSchema, //TODO change to TopicsSchema
-  rootValue: TopicsResolver,
-  graphiql: true,
+    schema: TopicsSchema, //TODO change to TopicsSchema
+    rootValue: TopicsResolver,
+    graphiql: true
 }));
 
 var server = http.createServer(app);
--- a/src/connectors/CassandraConnector.js
+++ b/src/connectors/CassandraConnector.js
@@ -6,40 +6,40 @@ const cassandra = Promise.promisifyAll(require('cassandra-driver'));
 const CASSANDRA_CONTACT_POINTS = process.env.CASSANDRA_CONTACT_POINTS;
 
 const options = {
-    contactPoints: [CASSANDRA_CONTACT_POINTS],
-    keyspace: process.env.CASSANDRA_KEYSPACE
+  contactPoints: [CASSANDRA_CONTACT_POINTS],
+  keyspace: process.env.CASSANDRA_KEYSPACE
 };
 
 module.exports = {
 
-    openClient: function () {
-        const client = new cassandra.Client(options);
+  openClient: function () {
+    const client = new cassandra.Client(options);
 
-        return new Promise((resolve, reject) => {
-            client.connect()
+    return new Promise((resolve, reject) => {
+      client.connect()
         .then(() => {
           //console.log('Connected to cluster with %d host(s): %j', client.hosts.length, client.hosts.keys);
           //console.log('Keyspaces: %j', Object.keys(client.metadata.keyspaces));
-            resolve(client);
+          resolve(client);
         })
         .catch((err) => {
           //console.error('There was an error when connecting to cluster with %d host(s): %j', client.hosts.length, client.hosts.keys, err);
-            client.shutdown();
-            reject(err);
+          client.shutdown();
+          reject(err);
         });
-        });
-    },
+    });
+  },
 
-    closeClient: function(client) {
-        return new Promise((resolve, reject) => {
-            client.shutdown()
+  closeClient: function(client) {
+    return new Promise((resolve, reject) => {
+      client.shutdown()
         .then(() => {
-            resolve();
+          resolve();
         })
         .catch((err) => {
           //console.error('There was an error shutting down the client');
-            reject(err);
+          reject(err);
         });
-        });
-    }
+    });
+  }
 };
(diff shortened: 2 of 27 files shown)


src/connectors/CassandraConnector.js
```diff
@@ -0,0 +1,45 @@
...
+
+module.exports = {
+
+  openClient: function () {
+    const client = new cassandra.Client(options);
+
+    return new Promise((resolve, reject) => {
+      client.connect()
+        .then(() => {
+          //console.log('Connected to cluster with %d host(s): %j', client.hosts.length, client.hosts.keys);
```
c-w: Please remove commented out code before merging.

src/connectors/CassandraConnector.js
```diff
@@ -0,0 +1,45 @@
...
+          client.shutdown();
+          reject(err);
+        });
+    });
+  },
+
+  closeClient: function(client) {
+    return new Promise((resolve, reject) => {
+      client.shutdown()
+        .then(() => {
```
c-w: Nit: could be simpler by skipping the lambda:

```js
client.shutdown()
.then(resolve)
.catch(reject);
```

server.js
```diff
@@ -16,21 +16,23 @@ var EdgesSchema = require('./src/schemas/EdgesSchema');
...
+var TopicsResolver = require('./src/resolvers/Topics');
 
 var app = express();
 
 app.use(function(req, res, next) {
-    res.header('Access-Control-Allow-Origin', '*');
-    res.header('Access-Control-Allow-Credentials', true);
-    res.header('Access-Control-Allow-Methods', 'OPTIONS,GET,POST,PUT,DELETE');
-    res.header('Access-Control-Allow-Headers', 'Content-Type, Authorization, x-ms-meta-data*,x-ms-meta-target*,x-ms-meta-abc');
+  res.header('Access-Control-Allow-Origin', '*');
```
c-w, line 25: For next time: this would have been easier to review as two pull requests, one for formatting/lint changes and one for functional change.

src/schemas/TopicsSchema.js
```diff
@@ -0,0 +1,19 @@
...
+    topic: String!,
+    value: String!
+  }
+
+  type TopicCollection {
+    collection: [Topic]!
+  }
+  
+  type Query {
+    get(containerName: String!, blobName: String!, id: Int!): Topic
```
c-w: This query interface is very tied to our data model (currently Azure Blob but may change in the future). Would it be reasonable to hide the Azure Blob details, e.g. by hardcoding the container name? We already use a single blob from where all Fortis users get their models, so might as well not require the user to know about that blob.

src/storageClients/BlobStorageManager.js
```diff
@@ -0,0 +1,89 @@
...
+module.exports = {
+
+  Get: function(containerName, blobName, id) {
+    let blobSvc = azure.createBlobService(BLOB_STORAGE_CONNECTION_STRING);
+    return new Promise((resolve, reject) => {
+      blobSvc.getBlobToTextAsync(containerName, blobName, null)
+        .then(text => {
+          let item = JSON.parse(text)[id];
+          if(item) {
+            resolve(JSON.parse(text)[id]);
```
c-w: Isn't this the same as `resolve(item)`?

src/storageClients/BlobStorageManager.js
```diff
@@ -0,0 +1,89 @@
...
+          } else {
+            resolve(null);
```
c-w: This requires the consuming code to null-check values on success. Would it make more sense to use reject here as essentially this branch represents a 404?

src/storageClients/BlobStorageManager.js
```diff
@@ -0,0 +1,89 @@
...
+const BLOB_STORAGE_CONNECTION_STRING = process.env.BLOB_STORAGE_CONNECTION_STRING;
+
+module.exports = {
+
+  Get: function(containerName, blobName, id) {
+    let blobSvc = azure.createBlobService(BLOB_STORAGE_CONNECTION_STRING);
+    return new Promise((resolve, reject) => {
+      blobSvc.getBlobToTextAsync(containerName, blobName, null)
+        .then(text => {
+          let item = JSON.parse(text)[id];
```
c-w: Do we need to handle malformed json and reject the promise on parse failure?

src/storageClients/BlobStorageManager.js
```diff
@@ -0,0 +1,89 @@
...
+          reject(err);
+        });
+    });
+  },
+
+  List: function(containerName, blobNames) {
+    return new Promise((resolve, reject) => {
+      Promise.all(getListPromises(containerName, blobNames))
+        .then(itemsArrays => {
+          let mergedItems = [].concat.apply([], itemsArrays);
```
c-w: Would this be more readable as a utility function with a name like `flatten`?

src/storageClients/BlobStorageManager.js
```diff
@@ -0,0 +1,89 @@
...
+          reject(err);
+        });
+    });
+  },
+
+  getBlobNamesWithSiteType: function(containerName, siteType) {
+    return new Promise((resolve, reject) => {
+      listBlobsInContainer(containerName)
+        .then(blobNames => {
+          blobNames = blobNames.filter(blobNames => (blobNames.indexOf(siteType) > -1));
```
c-w: Should the argument in the lambda be called `blobName` given that the collection is `blobNames`?

src/storageClients/BlobStorageManager.js
```diff
@@ -0,0 +1,89 @@
...
+        });
+    });
+  }
+
+};
+
+let getListPromises = (containerName, blobNames) => {
+  let blobSvc = azure.createBlobService(BLOB_STORAGE_CONNECTION_STRING);
+  let listPromises = [];
+  for ( let blobName of blobNames ) {
```
c-w: Would this be simpler using `blobNames.map`?

src/storageClients/CassandraTableStorageManager.js
```diff
@@ -0,0 +1,48 @@
+'use strict';
+
+const Promise = require('bluebird');
+
+const BATCH_LIMIT = 10;
+
+module.exports = {
+
+  prepareInsertTopic: function(topic) { //TODO: the insert statement to reflect the final topic schema
```
c-w: Elsewhere uses ES6 arrow functions -- might as well make this consistent, like `prepareInsertTopic: (topic) => ({ query: '...', params: [...] })`

src/storageClients/CassandraTableStorageManager.js
```diff
@@ -0,0 +1,48 @@
...
+    return {
+      query: 'INSERT INTO topics (id, topic, value) VALUES (?, ?, ?)',
+      params: [ topic.id, topic.topic, topic.value ]
+    };
+  },
+
+  batch: function(client, queries){
+    return new Promise((resolve, reject) => {
+      Promise.all(getBatchPromises(client, queries))
```
c-w: I wonder: would it also work to return `Promise.all(getBatchPromises(client, queries))` here instead of creating the wrapping promise with the no-op then/catch?

src/storageClients/CassandraTableStorageManager.js
```diff
@@ -0,0 +1,48 @@
...
+          reject(err);
+        });
+    });
+  }
+
+};
+
+let getBatchPromises = (client, queries) => {
+  let insertPromises = [];
+  for ( let i = 0, j = queries.length; i < j; i += BATCH_LIMIT ) {
```
c-w: Is this iterating over the array in chunks? If so: would it be more readable to extract a helper like in this StackOverflow [post](https://stackoverflow.com/a/10456644/3817588)?

c-w (approved):
LGTM. Some suggestions around readability and the query interface level of abstraction.

src/connectors/CassandraConnector.js
Smarker: true, thanks!

server.js
c-w: FYI: There's a really useful argument to `git add` with is `-p`. This will step you through all your changes one at a time and ask you if you want to stage them for commit. In this way, it's relatively easy to make targeted/one-purpose commits. In this specific case, you could have staged/committed all the lint changes first, pushed to a branch to make a PR, get through that PR (easy: it's all just whitespace changes so it'll take 5 mins to review and sign off) and then submit another PR for the functional change. Generally speaking, it's much easier to review code if it's only doing one thing: refactoring or adding functionality. If multiple purposes are mixed, it increases the mental tax on the reviewer's part as one has to be much more focused to not gloss over important parts.

src/schemas/TopicsSchema.js
Smarker: Yes, good point. I agree, if we move away from blob storage, we won't really care about the container name. Say we get info from Cassandra, we would just need the tablename and id. I will update this query to reflect that.

src/storageClients/BlobStorageManager.js
Smarker: Oh true, so I should break this into two steps. I think I forgot to do this when promisify-ing.

src/storageClients/BlobStorageManager.js
Smarker: yes, thanks I missed that!

src/storageClients/BlobStorageManager.js
Smarker: yea that makes sense, I originally thought that it shouldn't return an error since you just found nothing, but if you found nothing then you can't really go on in the process.

src/storageClients/BlobStorageManager.js
Smarker: ah yea that did cross my mind and then I forgot make something like that (since scala has one)!! that would make more sense!

src/storageClients/BlobStorageManager.js
Smarker: yep whoops

src/storageClients/BlobStorageManager.js
Smarker: yes that will look a lot nicer!

src/storageClients/CassandraTableStorageManager.js
Smarker: yes, that makes sense. I'm a little new to es6 arrow functions, so I think I accidentally forgot to make this into one...and possibly others. I'll change those as well if I find more.

src/storageClients/CassandraTableStorageManager.js
Smarker: I tried it, but it didn't pass my test :( So I don't think it works, unless I may have missed something.

src/storageClients/CassandraTableStorageManager.js
Smarker: yes, it is iterating over chunks of size equal to BATCH_LIMIT. I will look into the helper, too.

server.js
Smarker: ok sounds good I'll try to do that next time!

Smarker:
added **TOPIC_SEED_CONTAINER** to env, so I didn't have to include container as a parameter in the /get and /list

src/connectors/CassandraConnector.js
```diff
@@ -0,0 +1,37 @@
...
+const CASSANDRA_CONTACT_POINTS = process.env.CASSANDRA_CONTACT_POINTS;
+
+const options = {
+  contactPoints: [CASSANDRA_CONTACT_POINTS],
+  keyspace: process.env.CASSANDRA_KEYSPACE
+};
+
+module.exports = {
+
+  openClient: () => {
```
erikschlegel, line 15: I'm wondering if we can further generalize the cassandra driver to invoke a request and return the resultset in the promises `Resolve` function.

erikschlegel:
Going forward can we please split out general refactoring changes(eslint related) and new functionality(cassandra driver) into separate PR's and GH issues. Feel free to breakout broader sprint issues into more prescritive tasks.

src/resolvers/Topics.js
```diff
@@ -0,0 +1,21 @@
+'use strict';
+
+const Promise = require('bluebird');
```
erikschlegel, line 3: why are we using `bluebird` as opposed to vanilla require('promise')?

src/resolvers/Topics.js
```diff
@@ -0,0 +1,21 @@
...
+const blobStorageManager = require('../storageClients/BlobStorageManager');
+
+module.exports = {
+  get(args) {
```
erikschlegel, line 7: Can we add more general error handling below(i.e. app insights)

src/storageClients/BlobStorageManager.js
```diff
@@ -0,0 +1,75 @@
+'use strict';
+
+const Promise = require('bluebird');
+const azure = Promise.promisifyAll(require('azure-storage'));
```
erikschlegel, line 4: this seems like an expensive operation as the azure-storage library is a large one. I'd prefer to write the promise code out.

src/storageClients/BlobStorageManager.js
```diff
@@ -0,0 +1,75 @@
...
+
+const BLOB_STORAGE_CONNECTION_STRING = process.env.BLOB_STORAGE_CONNECTION_STRING;
+const TOPIC_SEED_CONTAINER = process.env.TOPIC_SEED_CONTAINER;
+
+module.exports = {
+
+  Get: (blobName, id) => {
+    let blobSvc = azure.createBlobService(BLOB_STORAGE_CONNECTION_STRING);
```
erikschlegel, line 12: You should check to make sure that `blobSvc` is defined before using it, and thread the error to app insights if it's not.

src/storageClients/BlobStorageManager.js
```diff
@@ -0,0 +1,75 @@
...
+
+const Promise = require('bluebird');
+const azure = Promise.promisifyAll(require('azure-storage'));
+
+const BLOB_STORAGE_CONNECTION_STRING = process.env.BLOB_STORAGE_CONNECTION_STRING;
+const TOPIC_SEED_CONTAINER = process.env.TOPIC_SEED_CONTAINER;
+
+module.exports = {
+
+  Get: (blobName, id) => {
```
erikschlegel, line 11: Can the name of this function be a bit more descriptive? Also, not sure what `id` is.

src/storageClients/BlobStorageManager.js
```diff
@@ -0,0 +1,75 @@
...
+    let blobSvc = azure.createBlobService(BLOB_STORAGE_CONNECTION_STRING);
+    return new Promise((resolve, reject) => {
+      blobSvc.getBlobToTextAsync(TOPIC_SEED_CONTAINER, blobName, null)
+        .then(JSON.parse)
+        .then(text => {
+          let item = text[id];
+          if (item) resolve(item);
+          else reject(null);
```
erikschlegel, line 19: You should provide more context on why this call would be rejected(i.e. `id not found`)

src/storageClients/BlobStorageManager.js
```diff
@@ -0,0 +1,75 @@
...
+        })
+        .catch(reject);
+    });
+  },
+
+  List: blobNames => {
```
erikschlegel, line 25: Can you be more descriptive with what `List` and `blobNames` represents.

src/storageClients/BlobStorageManager.js
```diff
@@ -0,0 +1,75 @@
...
+    return new Promise((resolve, reject) => {
+      Promise.all(getListPromises(blobNames))
+        .then(itemsArrays => {
+          resolve({'collection': flatten(itemsArrays)});
+        })
+        .catch(reject);
+    });
+  },
+
+  getBlobNamesWithSiteType: siteType => {
```
erikschlegel, line 35: is `BlobNames` synonymous with keywords? If yes, can you please rename.

src/storageClients/BlobStorageManager.js
```diff
@@ -0,0 +1,75 @@
...
+          let item = text[id];
+          if (item) resolve(item);
+          else reject(null);
+        })
+        .catch(reject);
+    });
+  },
+
+  List: blobNames => {
+    return new Promise((resolve, reject) => {
```
erikschlegel, line 26: This function is a bit difficult to follow. Can this be further simplified.

src/storageClients/CassandraTableStorageManager.js
```diff
@@ -0,0 +1,38 @@
+'use strict';
+
+const Promise = require('bluebird');
```
erikschlegel, line 3: Let's stick with traditional `Promises` unless there's features from bluebird that we're using.

src/storageClients/BlobStorageManager.js
```diff
@@ -0,0 +1,75 @@
...
+        })
+        .catch(reject);
+    });
+  }
+
+};
+
+let flatten = arrayOfArrays => [].concat.apply([], arrayOfArrays);
+
+let getListPromises = blobNames => {
```
erikschlegel, line 50: Lets chat tomorrow to generalize `getListPromises`

src/storageClients/BlobStorageManager.js
```diff
@@ -0,0 +1,75 @@
...
+          .then(JSON.parse)
+          .then(resolve)
+          .catch(reject);
+      })
+    );
+  });
+  return listPromises;
+};
+
+let listBlobsInContainer = () => {
```
erikschlegel, line 66: Lets chat tomorrow to generalize `listBlobsInContainer`

src/storageClients/CassandraTableStorageManager.js
```diff
@@ -0,0 +1,38 @@
...
+
+const BATCH_LIMIT = 10;
+
+module.exports = {
+
+  prepareInsertTopic: topic => { //TODO: the insert statement to reflect the final topic schema
+    return {
+      query: 'INSERT INTO topics (id, topic, value) VALUES (?, ?, ?)',
```
erikschlegel, line 11: Please update with the latest schema as this statement would fail.

erikschlegel (changes requested):

src/resolvers/Topics.js
Smarker: You're right, I don't need it here. I do need to include bluebird to use its promisifyAll. Files like settings.js and blobstoragemanager.js need bluebird.

src/resolvers/Topics.js
Smarker: I am going to create a wrapper for appinsights that we can use so that we can pull out logging from the business logic.

src/storageClients/BlobStorageManager.js
Smarker: True, there is a configuration you can add to apply promisifyAll certain functions rather than the whole library. I will add that config.

src/storageClients/BlobStorageManager.js
```diff
@@ -0,0 +1,75 @@
+'use strict';
+
+const Promise = require('bluebird');
+const azure = Promise.promisifyAll(require('azure-storage'));
+
+const BLOB_STORAGE_CONNECTION_STRING = process.env.BLOB_STORAGE_CONNECTION_STRING;
```
erikschlegel, line 6: We're going to have to connect to the blob container without using a connection string. The issue here is that the deployment will be run on azure subscriptions that will not have access to the PCT central storage container. So we're going to need to access the blob files via the `request` module.

erikschlegel (changes requested):

src/storageClients/CassandraTableStorageManager.js
Smarker: thanks, I removed it

src/storageClients/BlobStorageManager.js
Smarker: renamed Get to getTopic. `id` was just going to be the id/key of the topic.

src/storageClients/BlobStorageManager.js
Smarker: Renamed list to getTopicList. Renamed blobNames to blobPath where blobPath is something like 'siteTypes/humanitarian/topics/defaultTopics.json'

src/connectors/CassandraConnector.js
```diff
@@ -0,0 +1,37 @@
...
+const cassandra = Promise.promisifyAll(require('cassandra-driver'));
+
+const CASSANDRA_CONTACT_POINTS = process.env.CASSANDRA_CONTACT_POINTS;
+
+const options = {
+  contactPoints: [CASSANDRA_CONTACT_POINTS],
+  keyspace: process.env.CASSANDRA_KEYSPACE
+};
+
+module.exports = {
```
c-w, line 13: It looks as though this PR will take a while longer to get merged. As such, could you please split out the Cassandra-related functionality into a separate PR so that folks working on the resolvers can leverage this work?

PR #7, opened 2017-06-23
App insights

Smarker:
Added appinsights to my code from #6. Please let me know if there are any other areas where you think appinsights should be used or if you have suggestions for a different approach.

Added **APPINSIGHTS_INSTRUMENTATIONKEY** to env var.


src/storageClients/BlobStorageFactsManager.js
```diff
@@ -9,114 +9,114 @@ const FACTS_STORAGE_CONNECTION_STRING = process.env.FACTS_STORAGE_CONNECTION_STR
...
-        var blobSvc = azure.createBlobService(FACTS_STORAGE_CONNECTION_STRING);
-        let blobPrefix = id.split('-')[0] + '/' + id.split('-')[1] + '/' + id.split('-')[2];
-        let factIndex = id.split('-')[3];
-        blobSvc.listBlobsSegmentedWithPrefix(FACTS_CONTAINER_NAME, blobPrefix, null, null, (error, result, response) => { // eslint-disable-line no-unused-vars
-            if (error || !result || !result.entries || result.entries.length == 0) {
-                callback(null, error);
+  GetFact: function (id, callback, logger) { // eslint-disable-line no-unused-vars
+    var blobSvc = azure.createBlobService(FACTS_STORAGE_CONNECTION_STRING);
+    let blobPrefix = id.split('-')[0] + '/' + id.split('-')[1] + '/' + id.split('-')[2];
+    let factIndex = id.split('-')[3];
```
jcjimenez: I know you didn't change this, but do you know if we want to do something other than throw on a potential invalid index from a malformed id?

jcjimenez (approved):
LGTM

c-w:
Given that this is based on #6, I've temporarily changed the base of this review branch to #6's branch so that the changes in this pull request are made more obvious. There are a few ways forward from here:

1) After getting sufficient sign-offs, change the base to master again.
2) After getting a sign-off, merge the changes into #6.
3) Wait for #6 to be merged and then merge this.

src/connectors/CassandraConnector.js
```diff
@@ -16,22 +25,35 @@ module.exports = {
...
 
   closeClient: client => {
+    let start = Date.now();
     return new Promise((resolve, reject) => {
       client.shutdown()
-        .then(resolve)
-        .catch(reject);
+        .then(() => {
+          let duration = Date.now() - start;
+          iclient.trackMetric('Cassandra shutdown time', duration);
```
c-w: I wonder if we can make this more generic/re-usable. Pseudo-code:

```js
withMetric('Cassandra connection time', () => new Promise((resolve, reject) => {
  client.connect()
  .catch((err) => {
    client.shutdown();
	reject(err);
  });
});

function withMetric(metricName, promiseAction) {
  let start = Date.now();
  promiseAction().then((val) => {
    let duration = Date.now() - start;
	iclient.trackMetric(metricName, duration);
	resolve(val);
  })
  .catch((err) => {
    iclient.trackException(err)
	reject(err);
  });
}
```

Adding a slight level of abstraction on top of using the naked client also has the advantage of separating the logging implementation from the code that does the business logic. Here, everything is still fine, but it's a slippery slope and easy to get to the situation where the logging/argument-checking code takes up the majority of a method and drowns out the business logic parts. Using this decorator style to handle logging will set a good example to prevent that.

Maybe it's worth spending 15 minutes to see if we can factor out the logging code like this, but it's not a high priority item so don't worry if you run into a roadblock.

src/resolvers/Topics.js
```diff
@@ -1,21 +1,64 @@
 'use strict';
+const appInsights = require('applicationinsights');
+appInsights.setup(process.env.APPINSIGHTS_INSTRUMENTATIONKEY)
```
c-w: We have this code repeated in quite a few places. Would it be possible to encapsulate this in one place to reduce the duplication (e.g. `src/loggingClient`)?

This has the added benefit that we can introduce a null-object ApplicationInsights client if the instrumentation key is not set so that we can enable running the app locally without requiring the instrumentation key to be specified which is a better developer experience.

Pseudo-code:

```js
const ikey = process.env.APPINSIGHTS_INSTRUMENTATIONKEY

let client;
if (ikey) {
  appInsights.setup(ikey);
  client = appInsights.getClient();
} else {
  client = {
    trackException: (ex) => console.log("exception " + ex),
    trackEvent: (route, details) => console.log("event " + route + " " + JSON.stringify(details))
  };
}

module.exports = {
  client: client
};
```

c-w (approved):

src/connectors/CassandraConnector.js
```diff
@@ -1,10 +1,19 @@
 'use strict';
+const appInsights = require('applicationinsights');
+appInsights.setup(process.env.APPINSIGHTS_INSTRUMENTATIONKEY)
```
c-w: Please also document the key in the README.

erikschlegel: This is unnecessary as the  `applicationinsights` package automatically loads the instance based on the `APPINSIGHTS_INSTRUMENTATIONKEY` env variable https://github.com/Microsoft/ApplicationInsights-node.js#getting-started

src/connectors/CassandraConnector.js
```diff
@@ -1,10 +1,19 @@
...
+
+const iclient = appInsights.getClient();
 
 const Promise = require('bluebird');
 const cassandra = Promise.promisifyAll(require('cassandra-driver'));
 
 const CASSANDRA_CONTACT_POINTS = process.env.CASSANDRA_CONTACT_POINTS;
-
 const options = {
   contactPoints: [CASSANDRA_CONTACT_POINTS],
```
erikschlegel: You should check that the `CASSANDRA_CONTACT_POINTS ` env variable is defined before it's used, and raise an app insight exception.

src/connectors/CassandraConnector.js
Smarker: ok!

src/resolvers/Topics.js
Smarker: yes, good idea

src/connectors/CassandraConnector.js
```diff
@@ -1,10 +1,19 @@
 'use strict';
+const appInsights = require('applicationinsights');
+appInsights.setup(process.env.APPINSIGHTS_INSTRUMENTATIONKEY)
+  .setAutoDependencyCorrelation(true)
```
erikschlegel: I believe these all are enabled by default

src/connectors/CassandraConnector.js
Smarker: good idea, I will try to do this

Smarker:
I edited these files to address the comments for this PR:
* connectors/CassandraConnector -> executeQueries
* resolvers/Settings.js -> createOrReplaceSite (using insertSeedTopics)
* storageClients/BlobStorageManager -> getTopicsBySiteType
* storageClients/CassandraTableStorageManager -> prepareInsertTopic, batch (using chunkQueryPromises)

I addressed the comments by
* removing all instances of bluebird
* adding a connection pooling option to the Cassandra client
* making Cassandra client a singleton
* creating getTopicsBySiteType using a http request rather than BLOB_STORAGE_CONNECTION_STRING
* editing the insert topics statement
* adding app insights within the functions to track

src/connectors/CassandraConnector.js
```diff
@@ -1,37 +1,54 @@
...
+ * https://docs.datastax.com/en/developer/nodejs-driver/3.2/api/type.ClientOptions/
+ */
 const options = {
-  contactPoints: [CASSANDRA_CONTACT_POINTS],
-  keyspace: process.env.CASSANDRA_KEYSPACE
+  contactPoints: [process.env.CASSANDRA_CONTACT_POINTS],
+  keyspace: process.env.CASSANDRA_KEYSPACE,
+  pooling: {
+    coreConnectionsPerHost: {
+      [distance.local] : 10,
```
erikschlegel, line 17: How did you come up with these settings?

src/connectors/CassandraConnector.js
```diff
@@ -1,37 +1,54 @@
 'use strict';
+const appInsights = require('applicationinsights');
+appInsights.setup(process.env.APPINSIGHTS_INSTRUMENTATIONKEY).start();
```
erikschlegel, line 3: re: my earlier comment, setting `APPINSIGHTS_INSTRUMENTATIONKEY` is redundant as the `applicationinsights` library automatically does this for you

src/connectors/CassandraConnector.js
```diff
@@ -1,37 +1,54 @@
...
+const cassandra = require('cassandra-driver');
+const cassandraTableStorageManager = require('../storageClients/CassandraTableStorageManager');
+const distance = cassandra.types.distance;
 
-const Promise = require('bluebird');
-const cassandra = Promise.promisifyAll(require('cassandra-driver'));
-
-const CASSANDRA_CONTACT_POINTS = process.env.CASSANDRA_CONTACT_POINTS;
-
+/** Cassandra Client Options
```
erikschlegel, line 9: Can you comment your code a bit more so other developers have more context into how this driver works(i.e. connection pooling, etc)

src/connectors/CassandraConnector.js
```diff
@@ -1,37 +1,54 @@
...
-  openClient: () => {
-    const client = new cassandra.Client(options);
+module.exports = {
 
+  /** Execute a batch of queries
+   * @param Array<{query, params}> queries
+   * @param String eventIdentifier  A string to identify the event for appinsights
+   * @return {Promise}
+   */
+  executeQueries: (queries, eventIdentifier) => {
```
erikschlegel, line 35: I'm confused, why would one want to execute a batch of queries? Not sure if this is meant to address a batch of mutation statements(`insert`, `update`). If yes, please name your functions and variables in a way that helps other developers better understand your code.

src/connectors/CassandraConnector.js
```diff
@@ -1,37 +1,54 @@
...
+    if (!eventIdentifier || eventIdentifier === '') eventIdentifier = 'Cassandra batch';
     return new Promise((resolve, reject) => {
-      client.connect()
-        .then(() => {
-          resolve(client);
+      let numQueries = queries.length;
+      cassandraTableStorageManager.batch(client, queries)
+        .then(res => {
+          iclient.trackEvent(eventIdentifier, {numQueries: numQueries});
```
erikschlegel, line 41: I think this will be tracked in app insights a key value bag of `properties` and not `measurements`. Is that your intention here?

src/connectors/CassandraConnector.js
```diff
@@ -1,37 +1,54 @@
...
+          resolve(res);
```
erikschlegel, line 42: How does the connection get returned to the pool and/or closed?

src/connectors/CassandraConnector.js
```diff
@@ -1,37 +1,54 @@
...
+      [distance.remote] : 10
+    } 
+  }
 };
 
-module.exports = {
+/** Code should share the same Client instance across the application
+ * http://docs.datastax.com/en/developer/nodejs-driver/3.2/coding-rules/
+ */
+const client = new cassandra.Client(options);
```
erikschlegel, line 26: Can this throw an exception if none of the Contact points are available? Not sure if it does or not.

src/resolvers/Settings.js
```diff
@@ -31,23 +30,27 @@ module.exports = {
 
     return new Promise((resolve, reject) => {
```
erikschlegel, line 31: Where is `Promise` imported?

src/resolvers/Settings.js
```diff
@@ -31,23 +30,27 @@ module.exports = {
```
erikschlegel, line 30: Lets create a new resolver called `CreateNewSite` and rename `createOrReplaceSite`->`ReplaceSite` and refactor accordingly. Both resolvers have two very different code paths and the code quality is starting to suffer because of it.

src/storageClients/BlobStorageManager.js
```diff
@@ -1,75 +1,36 @@
...
+  */
+  getTopicsBySiteType: (siteType) => {
     return new Promise((resolve, reject) => {
-      listBlobsInContainer()
-        .then(blobNames => {
-          blobNames = blobNames.filter(blobName => (blobName.indexOf(siteType) > -1));
-          resolve(blobNames);
-        })
-        .catch(reject);
+      const url = `https://fortiscentral.blob.core.windows.net/settings/siteTypes/${siteType}/topics/defaultTopics.json`;
```
erikschlegel, line 15: the root context of the central host should be coming from the environment and not hard coded, as this is already available via the deployment.

src/resolvers/Settings.js
```diff
@@ -292,26 +295,24 @@ module.exports = {
...
-       })
-       .then(queries => {
-         return cassandraTableStorageManager.batch(client, queries);
-       })
-       .then(resolve)
-       .catch(reject);
+    blobStorageManager.getTopicsBySiteType(siteType)
+      .then(topics => {
+        let queries = [];
+        for(let topic of topics) {
```
erikschlegel, line 308: Im not sure I fully understand how this is working, as the json schema in the blob containers dont match up with how you're binding `topics` to the cassandra prepared statement ```{"id": 62, "topic": "climate", "value": "#climate"}, {"id": 63, "topic": "climate", "value": "climatechange"}, {"id": 64, "topic": "climate", "value": "globalwarming"}```

erikschlegel: Also, I dont see any unit tests in your PR, which I would assume would catch this as from the surface this looks like it would fail.

src/resolvers/Settings.js
erikschlegel: You'll also need to do some checks in the event the `CreateNewSite` is called on an already existing site, which in this case would continually add duplicate topics.

src/storageClients/BlobStorageManager.js
```diff
@@ -1,75 +1,36 @@
 'use strict';
-
-const Promise = require('bluebird');
-const azure = Promise.promisifyAll(require('azure-storage'));
-
-const BLOB_STORAGE_CONNECTION_STRING = process.env.BLOB_STORAGE_CONNECTION_STRING;
-const TOPIC_SEED_CONTAINER = process.env.TOPIC_SEED_CONTAINER;
+const appInsights = require('applicationinsights');
+appInsights.setup(process.env.APPINSIGHTS_INSTRUMENTATIONKEY).start();
```
erikschlegel, line 3: Please remove `APPINSIGHTS_INSTRUMENTATIONKEY` reference. https://github.com/Microsoft/ApplicationInsights-node.js#getting-started

src/storageClients/BlobStorageManager.js
erikschlegel, line 15: you should be performing some validation on `siteType`

src/storageClients/BlobStorageManager.js
```diff
@@ -1,75 +1,36 @@
...
+
+      const GET = {
+        url: url,
+        json: true,
+        withCredentials: false
+      };
+
+      request(GET, (error, response, body) => {
+        if (!error && response.statusCode === 200 && body) {
+          iclient.trackEvent(`GET ${url}`);
```
erikschlegel, line 25: the event name should be something thats short and consistent(`REQ_SITE`). You should be adding the url as a key entry in the `properties` map.

src/storageClients/BlobStorageManager.js
```diff
@@ -1,75 +1,36 @@
...
+          resolve(body);
+        } else {
+          const errMsg = `[${error}] occured while fetching request for url ${url}`;
```
erikschlegel, line 28: I'd also log the ` response.statusCode` in your `errMsg`

src/storageClients/CassandraTableStorageManager.js
```diff
@@ -1,15 +1,34 @@
 'use strict';
-
-const Promise = require('bluebird');
-
 const BATCH_LIMIT = 10;
 
 module.exports = {
 
-  prepareInsertTopic: topic => { //TODO: the insert statement to reflect the final topic schema
+  prepareInsertTopic: topic => {
```
erikschlegel, line 6: This insert statement doesnt belong in the table storage manager. Perhaps have a separate file to maintain your queries / mutation  statements.

src/storageClients/CassandraTableStorageManager.js
```diff
@@ -1,15 +1,34 @@
 'use strict';
-
-const Promise = require('bluebird');
-
 const BATCH_LIMIT = 10;
 
 module.exports = {
```
erikschlegel, line 5: It's a bit difficult to follow the `chunkQueryPromises ` method. I mentioned using the `async` forEachLimit` in an earlier comment. Would probably help simplify things a bit.

src/storageClients/CassandraTableStorageManager.js
```diff
@@ -1,15 +1,34 @@
...
+        topic.period,
+        topic.pipeline,
+        topic.sourceid,
+        topic.topic,
+        topic.topiclangcode,
+        topic.insertion_time,
+        topic.computedfeatures
+      ]
     };
   },
```
erikschlegel, line 34: the function name of `batch` seems a bit vague.

src/storageClients/CassandraTableStorageManager.js
```diff
@@ -1,15 +1,34 @@
...
 const BATCH_LIMIT = 10;
 
 module.exports = {
 
-  prepareInsertTopic: topic => { //TODO: the insert statement to reflect the final topic schema
+  prepareInsertTopic: topic => {
     return {
-      query: 'INSERT INTO topics (id, topic, value) VALUES (?, ?, ?)',
-      params: [ topic.id, topic.topic, topic.value ]
+      query: `INSERT INTO computedtopics (
```
erikschlegel, line 8: spark should be the only process writing to this table. I think you meant to write to `watchlist`

test/CassandraConnector.js
```diff
@@ -1,25 +1,28 @@
...
-        keyspace: process.env.CASSANDRA_KEYSPACE
-      };
-      const client = new cassandra.Client(options);
-      return (expect(cassandraConnector.closeClient(client)).to.eventually.be.fulfilled);
+  describe('#executeQueries(queries, eventIdentifier)', function() {
+    
+    it('should resolve the promise', function() {
+      let queries = [
+        {
+          query: 'INSERT INTO Topics (id, topic, value) VALUES (?, ?, ?)',
```
erikschlegel, line 15: there is no table called `Topics` https://github.com/CatalystCode/fortisdeploy/blob/119d3d847826b2cbd45791b3cf41c2ce1d60144d/ops/storage-ddls/cassandra-setup.cql

erikschlegel (changes requested):
I'd like to better understand how the data quality for the results of these changes are being validated and unit tested. 

Lets sync up when you have a minute so we can walk through the comments. Probably makes sense to only focus on this feature until we're at a point where it's ready for a production release.

PR #8, opened 2017-06-26
Always declare queries/mutations at top of schema

c-w:
No functional change, just some consistency fixes to make it easier to scan-read all the schemas.


Smarker (approved):
LGTM this makes sense

PR #10, opened 2017-06-26
Remove unused TopLocations schema and resolver

c-w:
As per [project-fortis-spark#17](https://github.com/CatalystCode/project-fortis-spark/issues/17), this schema is unused. We should remove it so that we don't re-implement it when we move to the Cassandra backend.


erikschlegel (approved):
LGTM

PR #13, opened 2017-06-28
Implement {save,remove}Locations resolver

c-w:
As mentioned in [project-fortis-spark#17](https://github.com/CatalystCode/project-fortis-spark/issues/17), we no longer need these endpoints as the functionality is now handled directly by Spark.

### DIFF
--- a/src/resolvers-cassandra/Edges.js
+++ b/src/resolvers-cassandra/Edges.js
@@ -1,7 +1,7 @@
 'use strict';
 
 module.exports = {
-    // ---------------------------------------------------------------------------------- mutations
+  // ---------------------------------------------------------------------------------- mutations
 
   removeKeywords(args, res){ // eslint-disable-line no-unused-vars
   },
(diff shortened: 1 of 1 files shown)


src/resolvers-cassandra/Edges.js
```diff
@@ -10,12 +10,26 @@ module.exports = {
   },
 
   saveLocations(args, res){ // eslint-disable-line no-unused-vars
+    return new Promise((resolve, reject) => { // eslint-disable-line no-unused-vars
+      reject(
+        'This API call is no longer supported. ' +
+        'We now automatically filter events down to only those ' +
+        'locations defined in the geo-fence for your site'
+      );
+    });
```
jcjimenez, line 19: Do we need to keep these just for backwards compatibility?

jcjimenez (approved):
LGTM with minor question.

src/resolvers-cassandra/Edges.js
Smarker: I was wondering that as well. Also, does saveLocations in /resolvers need to be changed, too?

c-w: Yes, the idea is to have the cassandra-based resolvers to be 100% backwards compatible with the Fortis-v1 UI. This has the advantage that once all of our Fortis-v2 backend work is done, we can just hook up the v2 backend to the v1 frontend for a first milestone and then focus on revamping the frontend as a second milestone without having to worry about an in-flux backend.

The methods in /resolvers don't have to be changed as those are the Fortis-v1 implementations which we're not touching. The GraphQL+Cassandra work is entirely localized to Fortis-v2.

PR #15, opened 2017-06-28
Split out client to send EventHub messages

c-w:
No functional change.


Smarker (approved):
LGTM, I like that you split this out.

PR #20, opened 2017-06-29
Detect bad imports in eslint

c-w:
After merging this pull request, the eslint step in our CI pipeline will fail if there is any `require` call against a dependency that doesn't exist.

For example, this will protect us against the following types of problems:

```js
const Promise = require('promse');
const eventHubSender = require('../eventhubClient/EventHubSemder');
```

Sample build that detected a bad import: [Job#94.1](https://travis-ci.org/CatalystCode/project-fortis-services/jobs/248471945).

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -22,7 +22,13 @@
         "es6": true,
         "node": true
     },
-    "extends": "eslint:recommended",
+    "extends": [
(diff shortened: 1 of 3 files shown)


Smarker (approved):
LGTM

PR #21, opened 2017-06-29
Introduce clients package

c-w:
This pull request improves the organization of our package structure by introducing a wrapper package for all clients. After merging, our folder structure will be much simpler: `schemas` and `resolvers` for GraphQL plus `clients` for third-party integrations.

While looking at all of our `require` statements, I also fixed a few inconsistencies like always using `const` and never relying on an implicit `Promise` object.


Smarker (approved):
LGTM

PR #22, opened 2017-06-29
Implement client for feature service

c-w:
This will be required for a few of the GraphQL resolvers, e.g. `MessageSchema.byBbox`, `TilesSchema.fetchPlacesByBbox` and `EdgesSchema.locations`.

~~Should be merged after #21.~~


Smarker (approved):
LGTM but maybe you'd want to put modules.export at the top so that we can easily see all the functions we have available to us?

src/clients/locations/FeatureServiceClient.js
```diff
@@ -0,0 +1,76 @@
...
+function formatBboxUri(north, west, south, east) {
+  return `http://${apiHost}/features/bbox/${north}/${west}/${south}/${east}`;
+}
+
+function formatPointUri(latitude, longitude) {
+  return `http://${apiHost}/features/point/${latitude}/${longitude}`;
+}
+
+function formatNameUri(names) {
+  return `http://${apiHost}/features/name/${names.join(',')}`;
```
jcjimenez: Do we need to URL escape all these values?

jcjimenez (approved):
LGTM with question.

c-w:
@Smarker Always declaring the exports at the bottom of the file also makes it easy to find them and has the benefit that it always works, even for exports that are declared as variables instead of functions ([read more](https://stackoverflow.com/q/23791749/3817588)).

src/clients/locations/FeatureServiceClient.js
c-w: Done in c381784.

Smarker:
Alright, good point. I agree.

PR #24, opened 2017-06-29
Cassandra connect

Smarker:
This is just the piece that let's you execute a batch of mutations. There are still more things that I need to think about and these are captured in the TODOs of the comments.


src/clients/cassandra/CassandraConnector.js
```diff
@@ -0,0 +1,79 @@
+'use strict';
+let Promise = require('promise');
```
c-w, line 2: s/let/const/

src/clients/cassandra/CassandraConnector.js
```diff
@@ -0,0 +1,79 @@
+'use strict';
```
c-w, line 1: Nit: elsewhere has newline after this statement.

src/clients/cassandra/CassandraConnector.js
```diff
@@ -0,0 +1,79 @@
...
+        }
+      });
+    });
+  },
+
+  /** Execute a query
+   * @param {Object} preparedQuery - cassandra prepared query object
+   * @param {String} preparedQuery.query - query string
+   * @param {Array}  preparedQuery.params - query params
+   * @return {Promise}
```
c-w, line 71: What's the return type of the Promise? Please document via `{Promise.<SomeType>}`

src/clients/cassandra/CassandraConnector.js
```diff
@@ -0,0 +1,79 @@
...
+          reject(errMsg);
+        } else {
+          resolve();
+        }
+      });
+    });
+  },
+
+  /** Execute a query
+   * @param {Object} preparedQuery - cassandra prepared query object
```
c-w, line 68: This and the next two lines can be simplified to `@param {{query: string, params: Array}} preparedQuery` which is less code (so easier to parse mentally) and immediately familiar to anyone who's written TypeScript.

src/clients/cassandra/CassandraConnector.js
```diff
@@ -0,0 +1,79 @@
...
+   * @param {String} preparedQuery.query - query string
+   * @param {Array}  preparedQuery.params - query params
```
c-w, line 70: What's the type of this array? If it's a primitive, you can use `string[]`, or for a complex type `Array<{foo: string, bar: number}>`.

src/clients/cassandra/CassandraConnector.js
```diff
@@ -0,0 +1,79 @@
...
+  /** Execute a batch of mutations
+   * @param Array[preparedMutation] mutations - cassandra preparedMutation array
+   * @param {Object} preparedMutation - cassandra prepared mutation object
+   * @param {String} preparedMutation.mutation - cassandra mutation string
+   * @param {Array}  preparedMutation.params - cassandra params
+   * @return {Promise}
+   */
+  executeBatchMutations: (mutations) => {
+    return new Promise((resolve, reject) => {
+      if(!client) reject('Cassandra client is null');
```
c-w, line 54: Would just remove this implementation for now if it's not yet finished.

src/clients/cassandra/CassandraConnector.js
```diff
@@ -0,0 +1,79 @@
...
+        } else {
+          resolve();
+        }
+      });
+    });
+  },
+
+  /** Execute a query
+   * @param {Object} preparedQuery - cassandra prepared query object
+   * @param {String} preparedQuery.query - query string
```
c-w, line 69: Here and elsewhere: no need to explicitly re-state what the variable does: that's already implied by choosing a good variable name and type.

src/clients/cassandra/CassandraConnector.js
```diff
@@ -0,0 +1,79 @@
...
+ * http://docs.datastax.com/en/developer/nodejs-driver/3.2/coding-rules/
+ * 
+ * cassandra.Client will fail if its client options property, contactPoints, is not set.
+ * 
+ * TODO: Look into client.shutdown().
+ * Does this close all the connections in the connection pool?
+ * If one of the connections has an exception, I don't 
+ * want shutdown to close all the connections in the pool.
+ */
+const client = new cassandra.Client(options);
```
c-w, line 41: Can just remove all the setup for now -- just the interfaces to this module will be enough.

src/clients/cassandra/CassandraConnector.js
```diff
@@ -0,0 +1,79 @@
...
+module.exports = {
+
+  /** Execute a batch of mutations
+   * @param Array[preparedMutation] mutations - cassandra preparedMutation array
+   * @param {Object} preparedMutation - cassandra prepared mutation object
+   * @param {String} preparedMutation.mutation - cassandra mutation string
+   * @param {Array}  preparedMutation.params - cassandra params
+   * @return {Promise}
+   */
+  executeBatchMutations: (mutations) => {
```
c-w, line 52: Would be cleaner to have these as separate functions so that it's more immediately obvious what the module exports:

```
function executeBatchMutations() { ... }
function executeQuery() { ... }

module.exports = {
  executeBatchMutations: executeBatchMutations,
  executeQuery: executeQuery
};
```

This is somewhat easier to read and separates the concerns of functionality from public interface declaration.

test/CassandraStatements.js
```diff
@@ -0,0 +1,39 @@
+let chai = require('chai');
```
c-w, line 1: If this PR is just adding interfaces, no need for the tests this time around.

src/statements/CassandraStatements.js
```diff
@@ -0,0 +1,34 @@
...
+  /** Create an INSERT prepared statement object.
+   * The prepared statement object has two properties, mutation and params.
+   * mutation: contains the topic INSERT statement
+   * params: contains the params to be binded to the INSERT statement
+   * 
+   * Preparing queries gets the best performance and will ensure that
+   * the JavaScript parameters are correctly mapped to Cassandra types.
+   * 
+   */
+  prepareInsertTopic: topic => {
```
c-w, line 16: Is it really worth to have a file with statements for every situation instead of just having the code that calls Cassandra provide their own statements? I don't see these queries/mutations being shared much across different resolvers, so having all of them in a single file just reduces locality of implementations, adds a file that will grow very large and attract merge conflicts.

Long story short: I'd leave the responsibility of making the correct statements up to the code that call the CassandraClient.

src/clients/storage/CassandraTableStorageManager.js
```diff
@@ -0,0 +1,38 @@
...
+let asyncEachLimit = require('async/eachLimit');
+
+/** http://docs.datastax.com/en/developer/nodejs-driver/3.2/features/batch/
+ * TODO: ASYNC_BATCH_LIMIT will need to be tuned.
+ * It is best to keep the batch size small (in the order of 10s).
+ * If the batch size is greater than 5k, then the server will issue a warning.
+ */
+const ASYNC_BATCH_LIMIT = 10;
+
+module.exports = {
```
c-w, line 11: In what way does this class add value on top of the CassandraClient? Would it make sense for the CassandraClient to abstract away/hide the batching logic so that it's easier for other code to use the client (i.e., they'd only have to know about one class, not multiples).

src/clients/cassandra/CassandraConnector.js
```diff
@@ -0,0 +1,79 @@
...
+    return new Promise((resolve, reject) => {
+      if(!client) reject('Cassandra client is null');
+      cassandraTableStorageManager.batchMutations(client, mutations, function(err) {
```
c-w, line 55: Can be an arrow function `(err) => { ... }`

c-w (changes requested):

PR #26, opened 2017-06-29
Cassandra connector interface

Smarker:


src/clients/cassandra/CassandraConnector.js
```diff
@@ -0,0 +1,24 @@
+'use strict';
+
+const Promise = require('promise');
+
+module.exports = {
+  executeBatchMutations: executeBatchMutations,
```
c-w: I'm a bit surprised that this export works given that the functions are declared as variables and therefore shouldn't be hoisted.

src/clients/cassandra/CassandraConnector.js
```diff
@@ -0,0 +1,24 @@
...
+  executeQuery: executeQuery  
+};
+
+/** Execute a batch of mutations
+ * @param {Array<{mutation: string, params: Array<string|number|map}>} mutations
+ * @returns {Promise}
+ */
+let executeBatchMutations = (mutations) => { // eslint-disable-line no-unused-vars
```
c-w: Consistency: rest of codebase declares functions just via `function`.

src/clients/cassandra/CassandraConnector.js
```diff
@@ -0,0 +1,24 @@
+'use strict';
+
+const Promise = require('promise');
+
+module.exports = {
```
c-w: Consistency: rest of codebase has the exports at the bottom.

c-w (approved):
LGTM.

src/clients/cassandra/CassandraConnector.js
Smarker: yes I was wondering if it would be better at the top so you can quickly glance and see all the functions that you would have available to you.

src/clients/cassandra/CassandraConnector.js
Smarker: oh true sorry, yes you are right

src/clients/cassandra/CassandraConnector.js
Smarker: alright, I will make the change

PR #28, opened 2017-06-30
Implement executeBatchMutations in CassandraConnector.js

Smarker:


src/clients/storage/CassandraTableStorageManager.js
```diff
@@ -0,0 +1,42 @@
+'use strict';
+
+const asyncEachLimit = require('async/eachLimit');
+
+/** http://docs.datastax.com/en/developer/nodejs-driver/3.2/features/batch/
+ * TODO: ASYNC_BATCH_LIMIT will need to be tuned.
+ * It is best to keep the batch size small (in the order of 10s).
+ * If the batch size is greater than 5k, then the server will issue a warning.
+ */
+const ASYNC_BATCH_LIMIT = 10;
```
jcjimenez: Not a deal-breaker for me, but I figured I'd ask: did we decide against using an environment variable for this and defaulting to 10 if it isn't present?

jcjimenez (approved):
LGTM

src/clients/cassandra/CassandraConnector.js
```diff
@@ -1,17 +1,75 @@
...
- * @param {Array<{mutation: string, params: Array<string|number|map}>} mutations
+ * @param {Array<{mutation: string, params: Array<string|map>}>} mutations
  * @returns {Promise}
  */
-function executeBatchMutations(mutations) { // eslint-disable-line no-unused-vars
-  return new Promise((resolve, reject) => { // eslint-disable-line no-unused-vars
-    
+function executeBatchMutations(mutations) {
+  return new Promise((resolve, reject) => {
+    if(!client) reject('Cassandra client is null');
```
c-w: Error message is not technically correct. `No Cassandra client defined` would be less potentially misleading.

c-w: Same below for mutations.

src/clients/cassandra/CassandraConnector.js
```diff
@@ -1,17 +1,75 @@
...
+    if(!mutations) reject('Mutations is null');
```
c-w: Given that this is an array type, should we also check for `mutations.length` to early-out when no mutations are provided?

src/clients/cassandra/CassandraConnector.js
c-w: Do we need to return after the reject so that the rest of the method doesn't execute? [More details](https://stackoverflow.com/a/38181613/3817588).

src/clients/cassandra/CassandraConnector.js
```diff
@@ -1,17 +1,75 @@
...
+  keyspace: process.env.CASSANDRA_KEYSPACE,
+  pooling: {
+    coreConnectionsPerHost: {
+      [distance.local]: CORE_CONNECTIONS_PER_HOST,
+      [distance.remote]: CORE_CONNECTIONS_PER_HOST
+    } 
+  }
+};
+
+/** Code should share the same Client instance across the application i.e.
```
c-w: Comment is implied by the fact that you're instantiating this as a const and wrapping it with a custom client interface. Would thus avoid the comment as comments, like code,
 are a maintenance/comprehension cost (feel free to put the details in the commit message or pull request message though so that folks can refer to it in the future if more details are needed). You can use `git rebase -i master` to amend commit messages that are in the past.

c-w: Same for the options comments above.

src/clients/cassandra/CassandraConnector.js
```diff
@@ -1,17 +1,75 @@
...
+ * the client should be a singleton.
+ * http://docs.datastax.com/en/developer/nodejs-driver/3.2/coding-rules/
+ * 
+ * Creating the cassandra client will fail if its client options property, contactPoints, is not set.
+ * 
+ * client.shutdown() shuts down all connections to all hosts.
+ */
+const client = new cassandra.Client(options);
 
 /** Execute a batch of mutations
```
c-w: Comment is implied by function name.

src/clients/cassandra/CassandraConnector.js
```diff
@@ -1,17 +1,75 @@
...
  * @returns {Promise}
  */
-function executeBatchMutations(mutations) { // eslint-disable-line no-unused-vars
-  return new Promise((resolve, reject) => { // eslint-disable-line no-unused-vars
-    
+function executeBatchMutations(mutations) {
+  return new Promise((resolve, reject) => {
+    if(!client) reject('Cassandra client is null');
+    if(!mutations) reject('Mutations is null');
+    cassandraTableStorageManager.batchMutations(client, mutations, (err, res) => {
```
c-w: This table storage manager should be inlined in this file. The Cassandra client needs to be a singleton and it's inaccessible outside of this file, so the table storage manager is tied to this file by definition. Splitting the manager out into a separate module means that it's re-usable which is not the case here so the separation just adds another layer of indirection when trying to understand the code.

Also, the name `manager` is a good indication that it's probably a superfluous level of abstraction ;) Weasel-words like that are usually a code smell. Just in case you haven't read it yet: [Clean Code](https://github.com/rmsadik/x/blob/master/CleanCode/Clean%20Code.pdf) chapter 2 has some good rules of thumb around naming.

src/clients/cassandra/CassandraConnector.js
```diff
@@ -1,17 +1,75 @@
...
+        reject(errMsg);
+      } else {
+        resolve(res);
+      }
+    });
   });
 }
 
+/**
+ * http://docs.datastax.com/en/developer/nodejs-driver/3.2/features/udfs/
```
c-w: Nit: the canonical way to introduce "see more" links in JSDoc is `@see`. [Reference](http://usejsdoc.org/tags-see.html).

src/clients/storage/CassandraTableStorageManager.js
```diff
@@ -0,0 +1,42 @@
...
+        console.log('Finished executing cassandra prepared statements');
+      }
+      callback(finalCBErr, mutations);
+    }
+  );
+}
+
+function processMutation(client, mutationBatch, asyncCB) {
+  /* cassandra client batches on arrays */
+  if(!(mutationBatch instanceof Array)) mutationBatch = [mutationBatch];
```
c-w: Best way to check if a variable is not an array in Javascript is `mutationBatch.constructor !== Array`. [More details](https://stackoverflow.com/a/26633883/3817588).

src/clients/storage/CassandraTableStorageManager.js
```diff
@@ -0,0 +1,42 @@
...
+  client.batch(mutationBatch, { prepare: true }, (err) => {
+    if(err) {
+      console.log(err, 'failed to process batch');
+      asyncCB('Failed to process mutation batch');
```
c-w: I'd expect a callback to be called on success, not failure. Maybe the variable needs to be renamed. `onError` seems like a reasonable name.

src/statements/CassandraStatements.js
```diff
@@ -0,0 +1,29 @@
...
+ * the JavaScript parameters are correctly mapped to Cassandra types.
+ * 
+ * http://docs.datastax.com/en/developer/nodejs-driver/3.2/features/batch/
+ * This will be used for the batch() method, so the object needs to have
+ * two properties: query and params
+ * 
+ * @param {{keyword: string, lang_code: string, ?translations: map<string,string>}}
+ * @returns {{mutation: string, params: Array<string|map<string,string>>}}
+ */
+function prepareInsertTopic(topic) {
```
c-w: See my [comment](https://github.com/CatalystCode/project-fortis-services/pull/24#discussion_r124914826) on a previous review: I don't think it makes sense to have a utility module for the prepared statements since most of them won't get re-used across resolvers.

src/statements/CassandraStatements.js
```diff
@@ -0,0 +1,29 @@
...
+  if(!topic.keyword || !topic.lang_code) return null;
```
c-w: General remark: you also may want to check for nullity of the wrapper object itself: `if (!topic || !topic.keyword || !topic.lang_code) return null;`

As a more general comment: note that the `!variable` syntax is a bit of a dangerous habit since it'll break for number values so often we prefer `variable != null` which checks just for `null` and `undefined`. Here, using `!variable` is fine though since you also want to disallow empty strings.

src/clients/storage/CassandraTableStorageManager.js
```diff
@@ -0,0 +1,42 @@
...
+ * If the batch size is greater than 5k, then the server will issue a warning.
+ */
+const ASYNC_BATCH_LIMIT = 10;
+
+/** Used to execute multiple INSERT, UPDATE, or DELETE statements
+ * @param {Client} client
+ * @param {Array<{mutation: string, params: Array}>} mutations
+ */
+function batchMutations(client, mutations, callback) {
+  asyncEachLimit(mutations, ASYNC_BATCH_LIMIT, (mutationBatch, asyncCB) => processMutation(client, mutationBatch, asyncCB),
```
c-w: Please avoid abbreviations. `asyncCallback` is easier to read and with auto-complete no more work to type :)

test/CassandraStatements.js
```diff
@@ -0,0 +1,46 @@
+const chai = require('chai');
+const assert = chai.assert;
+const cassandraStatements = require('../src/statements/CassandraStatements');
+
+describe('Tests for CassandraStatements.js', function() {
+
+  describe('#prepareInsertTopic(topic)', function() {
```
c-w: The code that this tests is trivial: just returning a constant string and unpacking an object into an array. As such, this is the type of test that doesn't add much value since it doesn't encode anything really valuable to the system under test.

Having too many tests is just as bad as having too few tests since more tests increases the cost of code changes/refactors and gives a false sense of security. Ideally, we should test at system/interface boundaries.

For Cassandra, a meaningful test would be: "given these inputs, does this resolver hit the DB with the correct query?" or "given this DB response, does the resolver produce the correct output?". Both of these tests cover scenarios, not implementations so if they pass, we can assume with some degree of confidence that the resolver works. Additionally, the tests will likely not have to change if the implementation of the resolver changes.

Basically: it's better to test contracts, not implementations.

src/clients/storage/CassandraTableStorageManager.js
Smarker: No we didnt decide against it. Can we talk about this on monday?

c-w (changes requested):

src/clients/storage/CassandraTableStorageManager.js
c-w: When in doubt, just add the environment variable with a sane default ;) It's easier to change the environment variable value (deis can do this for us) than to change the code (requires redeployment of the app).

src/clients/storage/CassandraTableStorageManager.js
erikschlegel: I dont see `asyncCB` invoked. I believe this would run continuously?

erikschlegel (changes requested):

src/clients/storage/CassandraTableStorageManager.js
Smarker: ok makes sense

src/clients/cassandra/CassandraConnector.js
Smarker: ok thanks I did not realize this

src/clients/cassandra/CassandraConnector.js
Smarker: yes, good point!

src/clients/cassandra/CassandraConnector.js
Smarker: I see, I agree, that's good to look out for. I'll also check out that reading.

Smarker:
To address the comments:
JC: ASYNC_BATCH_LIMIT is now an environment variable.
Erik: I added asyncCallback so that the function would not run continuously.
Clemens: I believe I addressed all your comments in the last commit I made.

src/clients/cassandra/CassandraConnector.js
```diff
@@ -1,24 +1,90 @@
...
+ */
+const distance = cassandra.types.distance;
+const CORE_CONNECTIONS_PER_HOST = 1;
+const options = {
+  contactPoints: [process.env.CASSANDRA_CONTACT_POINTS],
+  keyspace: process.env.CASSANDRA_KEYSPACE,
+  pooling: {
+    coreConnectionsPerHost: {
+      [distance.local]: CORE_CONNECTIONS_PER_HOST,
+      [distance.remote]: CORE_CONNECTIONS_PER_HOST
```
erikschlegel: do we always want the local and remote settings to be the same?

src/clients/cassandra/CassandraConnector.js
```diff
@@ -1,24 +1,90 @@
...
  */
-function executeBatchMutations(mutations) { // eslint-disable-line no-unused-vars
-  return new Promise((resolve, reject) => { // eslint-disable-line no-unused-vars
-    
+function executeMutations(mutations) {
+  return new Promise((resolve, reject) => {
+    if(!client) return reject('No Cassandra client defined');
+    if(!mutations || mutations.length == 0) return reject('No mutations defined');
+    asyncEachLimit(mutations, ASYNC_BATCH_LIMIT, (mutation, asyncCallback) => {
+      client.execute(mutation.query, mutation.params, { prepare: true }, (err) => {
```
erikschlegel: Shouldnt we be using `client.batch` for mutations?

src/clients/cassandra/CassandraConnector.js
```diff
@@ -1,24 +1,90 @@
...
+      }
+    });
   });
 }
 
+/**
+ * @see http://docs.datastax.com/en/developer/nodejs-driver/3.2/features/udfs/
+ */
 function executeQuery(query) { // eslint-disable-line no-unused-vars
   return new Promise((resolve, reject) => { // eslint-disable-line no-unused-vars
```
erikschlegel: why is this empty?

erikschlegel (changes requested):

src/clients/cassandra/CassandraConnector.js
Smarker: No, I don't think so. For now I have the CORE_CONNECTIONS_PER_HOST, so I think I will instead add CORE_CONNECTIONS_PER_HOST_LOCAL and CORE_CONNECTIONS_PER_HOST_REMOTE

src/clients/cassandra/CassandraConnector.js
```diff
@@ -1,18 +1,80 @@
...
+    coreConnectionsPerHost: {
+      [distance.local]: CORE_CONNECTIONS_PER_HOST,
+      [distance.remote]: CORE_CONNECTIONS_PER_HOST
+    } 
+  }
+};
+
+const client = new cassandra.Client(options);
+
+/** Execute mutations
```
c-w: Comment is implied by function name.

src/clients/cassandra/CassandraConnector.js
c-w: Reference link: http://docs.datastax.com/en/developer/nodejs-driver/3.2/features/batch/

src/clients/cassandra/CassandraConnector.js
```diff
@@ -1,18 +1,80 @@
...
+        if(err) {
+          console.log(err, `Mutation failed for ${JSON.stringify(mutation)}`);
+          asyncCallback(err);
+        } else {
+          asyncCallback();
+        }
+      });
+    },
+    (err) => {
+      if(err){
```
c-w: Nit: spacing `if (err) {`

c-w: Also for `else` below.

src/clients/cassandra/CassandraConnector.js
```diff
@@ -1,18 +1,80 @@
...
-/** Execute a batch of mutations
- * @param {Array<{mutation: string, params: Array<string|number|map}>} mutations
+/** @see http://docs.datastax.com/en/developer/nodejs-driver/3.2/features/batch/
+ * TODO: ASYNC_BATCH_LIMIT will need to be tuned.
+ * It is best to keep the batch size small (in the order of 10s).
+ * If the batch size is greater than 5k, then the server will issue a warning.
+ */
+const ASYNC_BATCH_LIMIT = process.env.ASYNC_BATCH_LIMIT || 10;
+
+/** Default Cassandra Client Options
```
c-w: We don't need any of these comments -- they're very useful now when you're implementing the functionality but they'll just add noise/unnecessary detail for future developers who just want to get a brief overview of what's going on. As such, the commit message or pull request text are better places to put these comments.

c-w (changes requested):

src/clients/cassandra/CassandraConnector.js
```diff
@@ -1,19 +1,60 @@
...
+ * @param {Array<{mutation: string, params: Array<string|map>}>} mutations
  * @returns {Promise}
  */
-function executeBatchMutations(mutations) { // eslint-disable-line no-unused-vars
-  return new Promise((resolve, reject) => { // eslint-disable-line no-unused-vars
-    
+function executeBatchMutations(mutations) {
+  return new Promise((resolve, reject) => {
+    if (!client) return reject('No Cassandra client defined');
+    if (!mutations || mutations.length == 0) return reject('No mutations defined');
```
c-w: Nit: `!mutations.length` to be consistent with elsewhere.

src/clients/cassandra/CassandraConnector.js
```diff
@@ -1,19 +1,60 @@
 'use strict';
 
 const Promise = require('promise');
+const cassandra = require('cassandra-driver');
+const asyncEachLimit = require('async/eachLimit');
+const array = require('lodash/array');
```
c-w: Wouldn't we usually call the imported lodash `_`?

src/clients/cassandra/CassandraConnector.js
```diff
@@ -1,19 +1,60 @@
...
+const BATCH_LIMIT = process.env.BATCH_LIMIT || 10;
```
c-w: This may be more immediately obvious named `MAX_OPERATIONS_PER_BATCH`.

src/clients/cassandra/CassandraConnector.js
```diff
@@ -1,19 +1,60 @@
...
+const ASYNC_OPERATIONS_LIMIT = process.env.ASYNC_OPERATIONS_LIMIT || 50;
```
c-w: This may be more immediately obvious named `MAX_CONCURRENT_BATCHES`

c-w (approved):
LGTM

src/clients/cassandra/CassandraConnector.js
Smarker: ok

src/clients/cassandra/CassandraConnector.js
Smarker: alright, I'll rename it

src/clients/cassandra/CassandraConnector.js
Smarker: true, good idea

src/clients/cassandra/CassandraConnector.js
Smarker: yes, I agree

src/clients/cassandra/CassandraConnector.js
```diff
@@ -1,19 +1,60 @@
 'use strict';
 
 const Promise = require('promise');
+const cassandra = require('cassandra-driver');
+const asyncEachLimit = require('async/eachLimit');
+const _ = require('lodash/array');
```
erikschlegel: you should only be importing the lodash function that you're using, as lodash is a massive library.

erikschlegel (changes requested):

src/clients/cassandra/CassandraConnector.js
c-w: PS: if all we need is just `_.chunk`, we might as well make our own utility for that since it can be done in ~5 lines. There's a good [post on SO](https://stackoverflow.com/a/10456644/3817588) that outlines a few approaches.

erikschlegel (approved):

src/clients/cassandra/CassandraConnector.js
```diff
@@ -1,19 +1,60 @@
 'use strict';
 
 const Promise = require('promise');
+const cassandra = require('cassandra-driver');
+const asyncEachLimit = require('async/eachLimit');
+const _ = require('lodash/chunk');
```
c-w, line 6: FYI: if we look at the [lodash sources](https://github.com/lodash/lodash/blob/e5e8f35c066c71a04ba584f65acc017d032c0174/chunk.js#L37), the chunk method is exported as a default. If this import is creating problems, you'll have to use `const chunk = require('lodash/chunk').default`. [More info](https://stackoverflow.com/a/35973567/3817588).

PR #29, opened 2017-07-03
Implement MessageSchema.event resolver

c-w:
[project-fortis#48](https://github.com/CatalystCode/project-fortis/issues/48)

### DIFF
--- a/src/clients/cassandra/CassandraConnector.js
+++ b/src/clients/cassandra/CassandraConnector.js
@@ -12,7 +12,12 @@ function executeBatchMutations(mutations) { // eslint-disable-line no-unused-var
   });
 }
 
-function executeQuery(query) { // eslint-disable-line no-unused-vars
+/**
+ * @param {string} query
+ * @param {string[]} params
+ * @returns {Promise.<object[]>}
+ */
+function executeQuery(query, params) { // eslint-disable-line no-unused-vars
   return new Promise((resolve, reject) => { // eslint-disable-line no-unused-vars
 
   });
--- a/src/resolvers-cassandra/Messages/queries.js
+++ b/src/resolvers-cassandra/Messages/queries.js
@@ -2,6 +2,8 @@
 
 const Promise = require('promise');
 const translatorService = require('../../clients/translator/MsftTranslator');
+const cassandraConnector = require('../../clients/cassandra/CassandraConnector');
+const featureServiceClient = require('../../clients/locations/FeatureServiceClient');
 
 /**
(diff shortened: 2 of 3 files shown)


src/clients/cassandra/CassandraConnector.js
```diff
@@ -12,7 +12,12 @@ function executeBatchMutations(mutations) { // eslint-disable-line no-unused-var
   });
 }
 
-function executeQuery(query) { // eslint-disable-line no-unused-vars
+/**
+ * @param {string} query
+ * @param {string[]} params
+ * @returns {Promise.<object[]>}
+ */
+function executeQuery(query, params) { // eslint-disable-line no-unused-vars
```
Smarker, line 20: I agree this needs 2 parameters rather than passing in an unnecessary object with query and params properties.

src/resolvers-cassandra/Messages/queries.js
```diff
@@ -29,6 +35,49 @@ function byEdges(args, res) { // eslint-disable-line no-unused-vars
  * @returns {Promise.<Feature>}
  */
 function event(args, res) { // eslint-disable-line no-unused-vars
+  const eventById = 'SELECT * FROM fortis.events WHERE id = ?';
+
+  return new Promise((resolve, reject) => {
+    const eventId = args && args.messageId;
+    if (!eventId) {
+      reject('No event id to fetch specified');
```
Smarker, line 43: If you would like I believe you can also just do `return reject('No event id to fetch specified');` and it should still prevent the below lines from executing.

src/resolvers-cassandra/Messages/queries.js
```diff
@@ -29,6 +35,49 @@ function byEdges(args, res) { // eslint-disable-line no-unused-vars
...
+            title: ev.title,
+            originalSources: ev.pipeline &&
+              [ev.pipeline],
+            language: ev.eventlangcode,
+            source: ev.sourceurl,
+            fullText: ev.messagebody
+          }
+        });
+      })
+      .catch(err => reject(err));
```
Smarker, line 77: I think you could do `catch(reject);` too right? The way you have it now is more explicit, so I like that way, too.

Smarker (approved):
LGTM with small comments

jcjimenez (approved):
LGTM

src/resolvers-cassandra/Messages/queries.js
c-w: I'll do this as a follow-up.

src/resolvers-cassandra/Messages/queries.js
c-w: I'll do this as a follow-up.

erikschlegel (commented):
LGTM

PR #30, opened 2017-07-03
Consistency fixes

c-w:
1. Always use `catch(reject)` for simple promise error handling.
2. Whitespace fixes
3. When rejecting a promise, inline the return.

No functional change.


Smarker (approved):
LGTM

PR #33, opened 2017-07-03
Implement SettingsSchema.sites.

jcjimenez:
graphql

### DIFF
--- a/src/resolvers-cassandra/Settings/queries.js
+++ b/src/resolvers-cassandra/Settings/queries.js
@@ -42,7 +42,7 @@ function sites(args, res) { // eslint-disable-line no-unused-vars
       if (rows.length > 1) return reject(`Got more than one site (got ${rows.length}) with id '${siteId}'`);
 
       const site = cassandraRowToSite(rows[0]);
-      resolve([site]);
+      resolve({sites:[site]});
     })
     .catch(reject)
     ;
--- a/src/resolvers-cassandra/Settings/queries.js
+++ b/src/resolvers-cassandra/Settings/queries.js
@@ -33,7 +33,7 @@ function sites(args, res) { // eslint-disable-line no-unused-vars
   return new Promise((resolve, reject) => {
     const siteId = args.siteId;
     if (!siteId) {
-      return reject("No site id to fetch specified");
+      return reject('No site id to fetch specified');
     }
 
     const siteById = 'SELECT * FROM fortis.sitesettings WHERE id = ?';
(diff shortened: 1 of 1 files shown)


src/resolvers-cassandra/Settings/queries.js
```diff
@@ -2,12 +2,58 @@
...
+
+    const siteById = 'SELECT * FROM fortis.sitesettings WHERE id = ?';
+    cassandraConnector.executeQuery(siteById, [siteId])
+    // .catch(err => reject(err))
+    .then(rows => {
+      if (rows.length < 1) {
+        return reject(`Could not find site with id ${siteId}`);
+      }
+      if (rows.length > 1) {
+        return reject(`Got more than one (${rows.length}) site with id ${siteId}`);
```
Smarker: This reject's statement might get confusing (suppose rows.length is 3 and siteId is 1):
`Got more than one 3 site with id 1`

src/resolvers-cassandra/Settings/queries.js
```diff
@@ -2,12 +2,58 @@
...
+      }
+
+      const site = cassandraRowToSite(rows[0]);
+      const siteCollection = Object.assign({}, {runTime: '' + (Date.now() - startTime), sites: [ site ]});
+      resolve(siteCollection);
+    })
+    .catch(err => reject(err))
```
Smarker: You can also do `catch(reject);` here.

Smarker (approved):
LGTM with comments

src/resolvers-cassandra/Settings/queries.js
jcjimenez: Good, point! But I think we should be safe since it will look more like this:

```
Got more than one (2) site with id d4d294ec-5b2c-4674-8f3f-e1eb7b8dbb47
```

jcjimenez: But I agree - let me work on making that more readable.

Smarker: Oh yea you are right! Didn't see the parentheses.

src/resolvers-cassandra/Settings/queries.js
```diff
@@ -2,12 +2,58 @@
...
+    if (!siteId) {
+      return reject('No site id to fetch specified');
+    }
+
+    const siteById = 'SELECT * FROM fortis.sitesettings WHERE id = ?';
+    cassandraConnector.executeQuery(siteById, [siteId])
+    // .catch(err => reject(err))
+    .then(rows => {
+      if (rows.length < 1) {
+        return reject(`Could not find site with id ${siteId}`);
```
c-w: Nit: could inline the return-reject like elsewhere `if (...) return reject(...)`

src/resolvers-cassandra/Settings/queries.js
```diff
@@ -2,12 +2,58 @@
...
+      }
+      if (rows.length > 1) {
+        return reject(`Got more than one (${rows.length}) site with id ${siteId}`);
+      }
+
+      const site = cassandraRowToSite(rows[0]);
+      const siteCollection = Object.assign({}, {runTime: '' + (Date.now() - startTime), sites: [ site ]});
```
c-w: This is implemented more generally in #34.

c-w (approved):
LGTM

PR #42, opened 2017-07-05
Implemented Settings.trustedTwitterAccounts.

jcjimenez:
trusted_twitter_sources

### DIFF
--- a/src/resolvers-cassandra/Settings/queries.js
+++ b/src/resolvers-cassandra/Settings/queries.js
@@ -54,11 +54,29 @@ function sites(args, res) { // eslint-disable-line no-unused-vars
 function twitterAccounts(args, res) { // eslint-disable-line no-unused-vars
 }
 
+function cassandraRowToTrustedTwitterAccount(row) {
+  return {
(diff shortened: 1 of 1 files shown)


src/resolvers-cassandra/Settings/queries.js
```diff
@@ -52,11 +52,29 @@ function sites(args, res) { // eslint-disable-line no-unused-vars
 function twitterAccounts(args, res) { // eslint-disable-line no-unused-vars
 }
 
+function cassandraRowToTrustedTwitterAccount(row) {
+  return {
+    RowKey: `${row.connector},${row.sourceid},${row.sourcetype}`,
```
jcjimenez, line 57: In the AzureTable world, this was the primary key of the database row, which IIRC was a UUID. In Cassandra, the key is a compound primary key made of of the `connector`, `sourceid`, and `sourcetype` columns:

```
CREATE TABLE fortis.trustedsources (
    connector text,
    sourceid text,
    sourcetype text,
    insertion_time timestamp,
    rank int,
    PRIMARY KEY (connector, sourceid, sourcetype)
...
```

Smarker (approved):
LGTM

PR #44, opened 2017-07-05
Allow environment to override http/https for apis

c-w:
As per standup today: environment variables should be able to override whether we talk to services via http or https.

### DIFF
--- a/src/clients/facebook/FacebookAnalyticsClient.js
+++ b/src/clients/facebook/FacebookAnalyticsClient.js
@@ -4,10 +4,10 @@ const Promise = require('promise');
 const request = require('request');
 
 const accessToken = process.env.FACEBOOK_AUTH_TOKEN;
-const apiHost = process.env.FACEBOOK_API_HOST || 'graph.facebook.com';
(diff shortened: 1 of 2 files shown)


src/clients/facebook/FacebookAnalyticsClient.js
```diff
@@ -4,10 +4,10 @@ const Promise = require('promise');
 const request = require('request');
 
 const accessToken = process.env.FACEBOOK_AUTH_TOKEN;
-const apiHost = process.env.FACEBOOK_API_HOST || 'graph.facebook.com';
+const apiHost = process.env.FACEBOOK_API_HOST || 'https://graph.facebook.com';
```
jcjimenez, line 7: The `host` part in the variable name seems misleading to me; may I suggest 'const baseUrl' maybe as a followup PR at some point?

jcjimenez (approved):
LGTM

src/clients/facebook/FacebookAnalyticsClient.js
c-w: Sounds good. Will make a follow up PR.

PR #46, opened 2017-07-05
Implement getTopicsBySiteType

Smarker:
`getTopicsBySiteType(siteType)` will be used in `resolvers/settings.js` to seed a site with topics.


src/clients/storage/BlobStorageClient.js
```diff
@@ -0,0 +1,39 @@
+'use strict';
+
+let request = require('request');
```
c-w: s/let/const

[What's the difference between let and const?](http://wesbos.com/let-vs-const/)

src/clients/storage/BlobStorageClient.js
```diff
@@ -0,0 +1,39 @@
...
+const apiUrlBase = process.env.FORTIS_CENTRAL_ASSETS_HOST || 'https://fortiscentral.blob.core.windows.net/';
```
c-w: Let's declare the default without trailing slash so that the url construction below is clearer.

c-w: Consistency: in the spark repo this environment variable is called `FORTIS_BLOB_HOST`.

src/clients/storage/BlobStorageClient.js
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * @param {string} siteType
```
c-w: TBH, I'd only include the type annotations for private functions if they're non-obvious: comments are code and code is maintenance/understanding burden. Here, the name encodes that you're returning a url which is necessarily a string in Javascript.

src/clients/storage/BlobStorageClient.js
```diff
@@ -0,0 +1,39 @@
...
+ */
+function buildTopicsSiteTypeUri(siteType) {
+  return `${apiUrlBase}settings/siteTypes/${siteType}/topics/defaultTopics.json`;
+}
+
+/**
+ * @param {string} siteType
+ * @returns {Promise.<Array<{keyword: string, lang_code: string, insertion_time: timestamp, translations: Map<string, string>}>>}
+*/
+function getTopicsBySiteType(siteType) {
```
c-w: The `get` prefix usually indicates something cheap, like a field access. Here we're doing an expensive operation (network request) so it's more obvious to call it `fetch`. Here's a good [SO thread](https://stackoverflow.com/q/2141818/3817588) on the topic.

test/BlobStorageClient.js
```diff
@@ -0,0 +1,36 @@
...
+
+const request = require('request');
+const sinon = require('sinon');
+const chai = require('chai');
+const blobStorageClient = require('../src/clients/storage/BlobStorageClient');
+chai.should();
+const chaiAsPromised = require('chai-as-promised');
+chai.use(chaiAsPromised);
+
+describe('Tests for BlobStorageClient.js', function() {
```
c-w, line 11: All `function() {` can be `() => {` for clarity/brevity.

c-w (approved):
LGTM.

src/clients/storage/BlobStorageClient.js
Smarker: whoops yea I forgot to make this a const

src/clients/storage/BlobStorageClient.js
Smarker: ok, the reason why I didn't do that was because of https://github.com/CatalystCode/fortisdeploy/blob/master/ops/setup-environment.sh#L22, but I'll look more into this

Smarker: After talking @c-w, said he would update `FORTIS_BLOB_HOST` in the scala code to `FORTIS_CENTRAL_ASSETS_HOST` to be consistent

src/clients/storage/BlobStorageClient.js
Smarker: sure sounds good

test/BlobStorageClient.js
Smarker: The reason why I used function here was because of http://mochajs.org/#arrow-functions
which stated: `Passing arrow functions (“lambdas”) to Mocha is discouraged. Due to the lexical binding of this, such functions are unable to access the Mocha context.`

src/clients/storage/BlobStorageClient.js
Smarker: ok good point

src/clients/storage/BlobStorageClient.js
```diff
@@ -0,0 +1,35 @@
...
+        return reject(`Unable to parse JSON for topics response ${body}: ${err}`);
+      }
+
+      resolve(topics);
+    });
+  });
+}
+
+module.exports = {
+  fetchTopicsBySiteType: fetchTopicsBySiteType
```
c-w, line 35: Just as a general remark: it's a bit odd to have a client API that does something very specific which will be used only in one place. I'd normally expect client APIs to provide generic functionality (e.g. "insert into cassandra" or "fetch from blob") and then the business logic (i.e., in our code base the resolvers) to build on top of that (e.g. "fetch the topics for a site type from blob").

In the interest of time, this is fine by me for now, but consider to refactor this when you implement the resolver so that the blobStorageClient exposes a "fetchJson" method which takes a blob path and then the resolver takes care of formulating the topics-specific path.

Smarker: ok will do

PR #47, opened 2017-07-05
add 'before' and 'after' to lint globals

Smarker:


c-w (approved):
LGTM, but check if we can just use an [eslint environment](http://eslint.org/docs/user-guide/configuring#specifying-environments) that includes all the testing globals instead of having to manually specify all of them.

Smarker:
Ok good idea, I'll do that next.

PR #54, opened 2017-07-05
Topics request

Smarker:
Refactor #46 to have `fetchJson(uri)` rather than `getTopicsBySiteType(siteType)`

Removed `chai-as-promised` since `chai` is sufficient.


c-w (approved):
LGTM.

PR #59, opened 2017-07-06
Cleanup server file

c-w:
No functional change, just:

1) Replacing `var` with `const`
2) Sorting imports
3) Make import naming consistent


jcjimenez (approved):
LGTM

PR #61, opened 2017-07-07
A few CQL fixes

c-w:
1) CQL does not have a general-purpose `OR` so replacing with `IN` where possible.
2) String literals need to be escaped with `'`, not `"`


Smarker (approved):
LGTM

jcjimenez (approved):
LGTM

PR #65, opened 2017-07-07
Rename mutation property in comment to query

Smarker:
No functional change. 

`executeBatchMutations` performs `client.batch`. `client.batch` accepts an array of `{query: string, params: Array<string|map>}` not `{mutation: string, params: Array<string|map>}`.


jcjimenez (approved):
LGTM

PR #67, opened 2017-07-07
Add switch to toggle Fortis-v2 mode

c-w:
Just set `ENABLE_V2=true` in your environment and you'll use the new resolvers by default.


jcjimenez (approved):
LGTM

PR #68, opened 2017-07-07
Remove all OR references

c-w:
Cassandra doesn't support disjunctive queries so we need to work around this. For now, executing multiple queries in parallel and joining the results in memory is the least time-intensive way to implement this. Going forward, we should make an audit of our query needs and review if our schema supports all of those operations in a more idiomatic way (e.g. Cassandra secondary indices won't scale!)


jcjimenez (approved):
LGTM

PR #69, opened 2017-07-07
Implemented Settings.{modifyFacebookPages, removeFacebookPages}

jcjimenez:
modify_facebook_pages

### DIFF
--- a/src/resolvers-cassandra/Settings/mutations.js
+++ b/src/resolvers-cassandra/Settings/mutations.js
@@ -48,7 +48,7 @@ function modifyFacebookPages(args, res) { // eslint-disable-line no-unused-vars
     const pages = args && args.input && args.input.pages;
     if (!pages || !pages.length) return reject('No pages specified');
     
-    const invalidPages = pages.filter(pages=>!pages.pageUrl);
+    const invalidPages = pages.filter(page=>!page.pageUrl);
     if (invalidPages.length > 0) {
       reject(`pageUrl required for ${JSON.stringify(invalidPages)}`);
       return;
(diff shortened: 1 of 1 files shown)


src/resolvers-cassandra/Settings/mutations.js
```diff
@@ -19,13 +20,64 @@ const TRUSTED_SOURCES_RANK_DEFAULT = 10;
...
  * @param {{input: {site: string, pages: Array<{pageUrl: string, RowKey: string}>}}} args
  * @returns {Promise.<{runTime: string, pages: Array<{pageUrl: string, RowKey: string}>}>}
  */
 function modifyFacebookPages(args, res) { // eslint-disable-line no-unused-vars
   return new Promise((resolve, reject) => {
-    reject('Modification not yet supported.');
+    const pages = args && args.input && args.input.pages;
+    if (!pages || !pages.length) return reject('No pages specified');
+    
+    const invalidPages = pages.filter(pages=>!pages.pageUrl);
```
c-w: FYI: Unless you want to provide the nice error message, you can also use `pages.some(page => !page.pageUrl)`

c-w: Maybe better to not over-write the pages variable in the filter scope. `pages.filter(page => !page.pageUrl)`

c-w (approved):
LGTM.

src/resolvers-cassandra/Settings/mutations.js
jcjimenez: Nice! Will try that.

jcjimenez: Good catch.

PR #73, opened 2017-07-07
Ensure that all Cassandra queries can run

c-w:
Implementation is terrible, but this is likely the best we can do with the current schema.

### DIFF
--- a/src/resolvers-cassandra/Edges/queries.js
+++ b/src/resolvers-cassandra/Edges/queries.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const Promise = require('promise');
+const flatten = require('lodash/flatten');
 const cassandraConnector = require('../../clients/cassandra/CassandraConnector');
 const featureServiceClient = require('../../clients/locations/FeatureServiceClient');
 const withRunTime = require('../shared').withRunTime;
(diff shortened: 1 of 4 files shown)


src/resolvers-cassandra/Edges/queries.js
```diff
@@ -1,39 +1,43 @@
...
+      clauses.push('(event_time <= ?)');
+      params.push(args.toDate);
+    }
+
+    if (pipeline) {
+      clauses.push('(pipeline = ?)');
+      params.push(pipeline);
+    }
+
+    let query = `SELECT detectedkeywords FROM fortis.events WHERE ${clauses.join(' AND ')} ALLOW FILTERING`;
```
jcjimenez, line 38: `ALLOW FILTERING` ouch!

src/resolvers-cassandra/Facts/queries.js
```diff
@@ -23,26 +24,24 @@ function cassandraRowToFact(row) {
...
 
-    clauses.push('(detectedkeywords CONTAINS ?)');
-    params.push(keyword);
+    if (keywordAndPipeline.b) {
+      clauses.push('(pipeline = ?)');
+      params.push(keywordAndPipeline.b);
+    }
 
-    const query = `SELECT * FROM fortis.events WHERE (${clauses.join(' AND ')})`;
+    const query = `SELECT * FROM fortis.events WHERE ${clauses.join(' AND ')} ALLOW FILTERING`;
```
jcjimenez, line 44: Another ouch. AFAIK, `ALLOW FILTERING` fetches everything and does driver-side filtering, right?

jcjimenez (approved):
I think this will work as long as we have a very small number of records.

src/resolvers-cassandra/Edges/queries.js
c-w: Yes, the schema right now requires us to do this :( We really have to formulate the v2 queries that we want and make sure that the DB schema supports them. Adding @erikschlegel for visibility.

src/resolvers-cassandra/Facts/queries.js
c-w: See [comment above](https://github.com/CatalystCode/project-fortis-services/pull/73#discussion_r126229285).

PR #74, opened 2017-07-07
Update translator auth to just use account key

c-w:
This has the advantage that the key is very easy to pass-through from the Azure ARM deployment when we set up the translation Cognitive Service.

Also, the documentation mentions the /sts/v1.0/issueToken as the standard way to authenticate with the translation service going forward, so this will be more future proof, see the [docs](http://docs.microsofttranslator.com/text-translate.html#Authorization).

Proof that the new auth works:

image


jcjimenez (approved):
LGTM

PR #76, opened 2017-07-10
Avoid Cassandra IN clause

c-w:
Separate point queries are usually faster.


c-w:
No longer necessary given recent changes in master.

PR #77, opened 2017-07-10
CreateSite and RemoveSite with appinsights

Smarker:


src/resolvers-cassandra/Settings/mutations.js
```diff
@@ -3,21 +3,106 @@
...
+  });
+}
+
+/**
+ * @param {{input: {siteType: string, targetBbox: number[], defaultZoomLevel: number, logo: string, title: string, name: string, defaultLocation: number[], storageConnectionString: string, featuresConnectionString: string, mapzenApiKey: string, fbToken: string, supportedLanguages: string[]}}} args
+ * @returns {Promise}
+ */
+function removeSite(args, res) { // eslint-disable-line no-unused-vars
+  return new Promise((resolve, reject) => {
+    cassandraConnector.executeBatchMutations([{
```
c-w, line 122: I wonder if it's possible to just return the `cassandraConnector.executeBatchMutations` here instead of wrapping it in a promise? The cassandraConnector already returns a promise and we're not doing any post-processing, so I think this code should be equivalent :)

```
function removeSite(args, res) {
  return cassandraConnector.executeBatchMutations([{...}]);
}
```

c-w (commented):
LGTM.

src/resolvers-cassandra/Settings/mutations.js
Smarker: yes true good point

src/resolvers-cassandra/Settings/index.js
```diff
@@ -4,7 +4,8 @@ const mutations = require('./mutations');
 const queries = require('./queries');
 
 module.exports = {
-  createOrReplaceSite: mutations.createOrReplaceSite,
+  createSite: mutations.createSite,
+  removeSite: mutations.removeSite,
```
jcjimenez, line 8: This may not be a big deal, but I thought I'd call it out: this is a non-backwards compatible change from the point of view of the non-cassandra resolver. Should we just add the two new ones and have the `createOrReplaceSite` always call `reject`? Also, should we add the `createSite` and `removeSite` to the old non-cassandra side?

jcjimenez (approved):
LGTM with question regarding backwards compatibility.

src/resolvers-cassandra/Settings/index.js
Smarker: Done in [b34b9cf](https://github.com/CatalystCode/project-fortis-services/pull/77/commits/b34b9cf10993e2f8ae6c1eecc15ddf20d3aecab2)

Smarker:
Need to still address comment here: https://github.com/CatalystCode/project-fortis-spark/pull/36/files
* add a record into the streams table upon a called to CreateOrReplaceSite, as long as the eventhub environment is properly setup.

.vscode/launch.json
```diff
@@ -0,0 +1,25 @@
+{
```
jcjimenez: Can I suggest we remove this file and add `.vscode/` to our .gitignore?

src/resolvers-cassandra/Settings/mutations.js
```diff
@@ -3,21 +3,151 @@
...
 
 const TRUSTED_SOURCES_CONNECTOR_TWITTER = 'Twitter';
 const TRUSTED_SOURCES_CONNECTOR_FACEBOOK = 'FacebookPage';
 const TRUSTED_SOURCES_RANK_DEFAULT = 10;
 
+function createOrReplaceSite(args, res) { // eslint-disable-line no-unused-vars
+  return new Promise((resolve, reject) => { // eslint-disable-line no-unused-vars
+    return reject(
+      'This API call is no longer supported. ' +
+      'We separated this functionality into createSite and removeSite '
```
jcjimenez: Can we tweak this comment to read something like "Its functionality has been separated into createSite and removeSite."?

src/resolvers-cassandra/Settings/mutations.js
```diff
@@ -3,21 +3,151 @@
...
+                  mapzenApiKey: '',
+                  fbToken: '',
+                  supportedLanguages:args.input.supportedLanguages,
+                  name: args.input.name
+                }
+              });
+            })
+            .catch(reject);
+        } 
+        if (rows.length == 1) return reject(`Site with sitename ${args.input.name} already exists.`);
```
jcjimenez: I have a feeling this `if` is meant to be an `else if` as part of what looks like a 3-element `if/else` chain. I say this because it looks like the `else` below will always be called alongside the insert above.

jcjimenez (changes requested):
Discussed offline.

.vscode/launch.json
Smarker: yes definitely

src/resolvers-cassandra/Settings/mutations.js
Smarker: ok

jcjimenez (approved):
Looks good assuming merge from JSON serialization fix does the trick.

.vscode/settings.json
```diff
@@ -1,5 +0,0 @@
-{
```
c-w, line 1: Why remove this file? It makes sure that we all have the same editor settings (which pass eslint).

Smarker:
`createSite` with defined `siteType`
image

`createSite` with undefined `siteType`
image

`createSite` with a duplicate site:
image

`removeSite` with an existing/nonexisting site
image

src/resolvers-cassandra/Settings/mutations.js
Smarker: I kept it in a promise, since I wanted to return an object in the resolve.

Smarker:
Graphql doesn't support overriding http codes (https://github.com/graphql/express-graphql/issues/111). As long as the decorators like `trackEvent` behave then the messages we have in the `resolve`/`reject` should be fine.

Smarker:
Still need to address: '[add a record](https://github.com/CatalystCode/project-fortis-spark/pull/36/files) into the streams table upon a called to CreateOrReplaceSite, as long as the eventhub environment is properly setup.', but I am waiting on [Cassandra Table Overhaul](https://github.com/CatalystCode/project-fortis-pipeline/pull/84/files/dd444ccaf0224954777af1474cc462cb99d6776f#diff-e79522b8c1a9d91a3f8005389e6302c2)

c-w (approved):
LGTM.

.vscode/settings.json
jcjimenez: I suggested we remove it and add it to .gitignore =P This was prompted by some leaked credentials from a previous commit (that, incidentally, were different from mine).

I'm not against adding it back as long as we guard against things like that :)

c-w: Just use `git add -p` to stage files and never face issues like this :P

c-w: Alternatively, if you let me know what the credentials look like, I can whip up a quick pre-commit hook to prevent checkins of credentials.

jcjimenez: It was basically an object referenced by the `env` key:

```
      "env":{
          /* leaked credentials here */
       }
```

So maybe we just make the pre-commit hook ensure there's no `env` in .vscode/settings.json?

Smarker: yes sorry my mistake, as soon as I realized this I immediately removed them.

src/clients/cassandra/CassandraConnector.js
```diff
@@ -65,13 +62,14 @@ function executeBatchMutations(mutations) {
  */
 function executeQuery(query, params) { // eslint-disable-line no-unused-vars
```
erikschlegel: How is paging supported for larger datasets? https://github.com/datastax/nodejs-driver#paging Should support a `fetchSize` parameter as opposed to the default of 5K records.

erikschlegel (changes requested):
Very close to completion.

src/clients/cassandra/CassandraConnector.js
Smarker: @jcjimenez implemented this part, but I can look into paging for this and update it accordingly

Smarker:
**Query without pageState**
image

**Query with pageState:**
image

Paging was only done on `twitterAccounts` to get initial feedback before making tons of graphql schema changes.

src/clients/cassandra/CassandraConnector.js
Smarker: Paging done with `pageSize` parameter in [9f40f2b](https://github.com/CatalystCode/project-fortis-services/pull/77/commits/9f40f2be88c7aad928595fa35c6373004f00f221)

src/schemas/SettingsSchema.js
```diff
@@ -3,14 +3,16 @@ const graphql = require('graphql');
 module.exports = graphql.buildSchema(`
   type Query {
     sites(siteId: String): SiteCollection
-    twitterAccounts(siteId: String!): TwitterAccountCollection
+    twitterAccounts(siteId: String!, pageState: String, pageSize: Int): TwitterAccountConnection
```
jcjimenez: Oh, this change in return type feels non-backwards, compatible to me. Why not add optional paging info to the `TwitterAccountCollection` type?

src/schemas/SettingsSchema.js
```diff
@@ -45,9 +47,18 @@ module.exports = graphql.buildSchema(`
     sites: [Site]!,
   }
 
+  type PageInfo {
+    cursor: String
+  }
+
   type TwitterAccountCollection {
+    accounts: [TwitterAccount]
```
jcjimenez: I propose we just add `pageInfo: PageInfo` here. As long as it's not required, we should remain backwards-compatible with Fortis v1 UI.

src/schemas/SettingsSchema.js
Smarker: The reason why I didn't add paging to the `TwitterAccountCollection` type was because I read in the [docs](http://graphql.org/learn/pagination/): `That leads us to a problem; though; how do we get the cursor from the object? We wouldn't want cursor to live on the User type; it's a property of the connection, not of the object.`

I agree with you in that it would be best to not have a non-backwards compatible change, so I'll update the schema!

Smarker:
`TwitterAccounts` is back to normal and uses `executeQuery`

image

Smarker:
`executeQuery` with `stream` can return 30k records in 759ms.

erikschlegel:
LGTM. Nice work

erikschlegel (approved):

PR #79, opened 2017-07-10
Remove data that's never passed from the client

c-w:
@erikschlegel: As we talked about, there is a bunch of data that is included in our GraphQL schema but that's never passed from the client to the server. I've removed all such data that I could find via visual inspection of the [project-fortis-interfaces](https://github.com/CatalystCode/project-fortis-interfaces) project so that our query requirements are made clearer.


jcjimenez (approved):
LGTM

Smarker (approved):
LGTM

PR #81, opened 2017-07-11
Correctly implement sourceFilter

c-w:
No sources passed means "select from all sources", not "select no sources". See [client code](https://github.com/CatalystCode/project-fortis-interfaces/blob/cf241d64fec65875682eb0978fcae1d715db6472/src/actions/Actions.js#L278) for reference.


jcjimenez (approved):
LGTM - I have a similar PR about this.

PR #82, opened 2017-07-11
Extract shared multi-queries code

c-w:
No functional change.

### DIFF
--- a/src/clients/cassandra/CassandraConnector.js
+++ b/src/clients/cassandra/CassandraConnector.js
@@ -4,6 +4,7 @@ const Promise = require('promise');
 const cassandra = require('cassandra-driver');
 const asyncEachLimit = require('async/eachLimit');
 const chunk = require('lodash/chunk');
+const flatten = require('lodash/flatten');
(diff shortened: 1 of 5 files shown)


src/clients/cassandra/CassandraConnector.js
```diff
@@ -79,7 +81,23 @@ function executeQuery(query, params) { // eslint-disable-line no-unused-vars
   });
 }
 
+/**
+ * @param {Array<{query: string, params: string[]}} queries
+ * @returns {Promise.<object[]>}
+ */
+function executeQueries(queries) {
+  return new Promise((resolve, reject) => {
+    Promise.all(queries.map(query => executeQuery(query.query, query.params)))
```
Smarker, line 90: What about using `asyncEachLimit` instead of `Promise.all`?

c-w: That'll be a good future enhancement.

Smarker: nice ok!

Smarker (approved):
LGTM, and it would be nice to add `asyncEachLimit` as a future enhancement.

PR #83, opened 2017-07-11
Corrected Twitter and Facebook connector references.

jcjimenez:


src/resolvers-cassandra/Settings/queries.js
```diff
@@ -6,6 +6,9 @@ const cassandraConnector = require('../../clients/cassandra/CassandraConnector')
 const withRunTime = require('../shared').withRunTime;
 const trackEvent = require('../../clients/appinsights/AppInsightsClient').trackEvent;
 
+const CONNECTOR_TWITTER = 'Twitter'
```
c-w: Shouldn't these be lowercase like the stream names? [Example](https://github.com/CatalystCode/project-fortis-spark/blob/5c324a98eceabedf58b84dcadaa7b13c712830e9/src/main/scala/com/microsoft/partnercatalyst/fortis/spark/ProjectFortis.scala#L178).

src/resolvers-cassandra/Settings/queries.js
```diff
@@ -6,6 +6,9 @@ const cassandraConnector = require('../../clients/cassandra/CassandraConnector')
...
+const CONNECTOR_FACEBOOK = 'Facebook'
```
c-w: Might as well move these constants to [shared.js](https://github.com/CatalystCode/project-fortis-services/blob/4846f58b2229a99b4ee7cd6d8bd47523313eccee/src/resolvers-cassandra/shared.js#L34) so that all the pipeline names are in one place.

c-w (approved):
LGTM.

jcjimenez:
Thanks, @c-w !

PR #85, opened 2017-07-12
Fix circular reference

c-w:
This fixes a circular reference when using the decorators on resolvers.


Smarker (approved):
LGTM - thanks!

jcjimenez (approved):
LGTM

PR #86, opened 2017-07-12
Adding extra guard for possible exceptions coming from exception handling logic.

jcjimenez:


src/clients/appinsights/AppInsightsClient.js
```diff
@@ -85,7 +85,9 @@ function trackEvent(promiseFunc, eventName, extraPropsFunc) {
         }
         console.error(JSON.stringify({event: eventName, properties: props, err, args: args && args.length && args[0]}));
         reject(err);
-      });
+      })
+      .catch(reject)
```
c-w: Looks good. Let's also add this to the other decorator.

c-w (approved):
LGTM.

PR #87, opened 2017-07-31
Update the GraphQL queries with new Cassandra schema

c-w:
Just to get us started, still some things to implement (e.g. extra data needs to get wired from the frontend).

### DIFF
--- a/src/resolvers-cassandra/Edges/queries.js
+++ b/src/resolvers-cassandra/Edges/queries.js
@@ -13,29 +13,17 @@ const { trackEvent } = require('../../clients/appinsights/AppInsightsClient');
  */
 function terms(args, res) { // eslint-disable-line no-unused-vars
   return new Promise((resolve, reject) => {
-    if (!args) return reject('No args specified');
-
-    const { fromDate, toDate } = parseFromToDate(args.fromDate, args.toDate);
-
     const query = `
-    SELECT topics
-    FROM fortis.mentionedtopics
-    WHERE pipelinekey = ?
-    AND externalsourceid = ?
-    AND event_time <= ?
-    AND event_time >= ?
+    SELECT topic
+    FROM fortis.watchlist
     `.trim();
 
     const params = [
-      toPipelineKey(args.sourceFilter),
-      'all',
-      toDate,
-      fromDate
     ];
 
     cassandraConnector.executeQuery(query, params)
     .then(rows => {
-      const keywords = makeSet(rows.map(row => row.topics), topic => topic);
+      const keywords = makeSet(rows, row => row.topic);
 
       return {
         edges: keywords.map(keyword => ({name: keyword}))
--- a/src/resolvers-cassandra/Edges/queries.js
+++ b/src/resolvers-cassandra/Edges/queries.js
@@ -123,14 +123,12 @@ function popularLocations(args, res) { // eslint-disable-line no-unused-vars
 }
 
 /**
- * @param {{site: string, fromDate: string, toDate: string, sourceFilter: string[], mainEdge: string}} args
+ * @param {{site: string, fromDate: string, toDate: string, sourceFilter: string[], mainEdge: string, bbox: number[], zoomLevel: number}} args
  * @returns {Promise.<{labels: Array<{name: string, mentions: number}>, graphData: Array<{date: string, edges: string[], mentions: number[]}>}>}
  */
 function timeSeries(args, res) { // eslint-disable-line no-unused-vars
   return new Promise((resolve, reject) => {
     const { period, periodType, fromDate, toDate } = parseFromToDate(args.fromDate, args.toDate);
-    // FIXME no bbox available
-    // FIXME no zoomLevel available
     const tiles = tilesForBbox(args.bbox, args.zoomLevel);
     const tilex = makeSet(tiles, tile => tile.row);
     const tiley = makeSet(tiles, tile => tile.column);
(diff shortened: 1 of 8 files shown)


src/resolvers-cassandra/Edges/queries.js
```diff
@@ -70,45 +53,182 @@ function locations(args, res) { // eslint-disable-line no-unused-vars
...
+    const { period, periodType } = parseTimespan(args.timespan);
+
+    const query = `
+    SELECT placename, placeid, mentioncount
+    FROM fortis.popularplaces
+    WHERE period = ?
+    AND periodtype = ?
+    AND pipelinekey = ?
+    AND externalsourceid = ?
+    AND conjunctiontopics = ?
```
erikschlegel: we've switched `conjunctivetopic1, conjunctivetopic2, conjunctivetopic3`

src/resolvers-cassandra/Edges/queries.js
```diff
@@ -13,29 +13,17 @@ const { trackEvent } = require('../../clients/appinsights/AppInsightsClient');
...
-      toPipelineKey(args.sourceFilter),
-      'all',
-      toDate,
-      fromDate
     ];
 
     cassandraConnector.executeQuery(query, params)
     .then(rows => {
-      const keywords = makeSet(rows.map(row => row.topics), topic => topic);
+      const keywords = makeSet(rows, row => row.topic);
```
Smarker: Rather than `const keywords` maybe it would be clearer to use `const topics` so there's one term for topic rather than a topic's name either being `keyword` or `topic`?

Smarker (approved):
LGTM

src/resolvers-cassandra/Messages/queries.js
```diff
@@ -130,21 +48,55 @@ function byBbox(args, res) { // eslint-disable-line no-unused-vars
...
+      const tagsQuery = `
+      SELECT eventids
+      FROM fortis.eventplaces
+      WHERE placeid IN ?
+      AND conjunctiontopics = ?
+      AND pipelinekey = ?
+      AND externalsourceid = ?
+      AND event_time <= ?
+      AND event_time >= ?
+      `.trim();
```
erikschlegel: we should probably use `LIMIT` here, as this result set can be large. We only care about the events that display on the news feed which are pages of 15 events.

src/resolvers-cassandra/Messages/queries.js
```diff
@@ -130,21 +48,55 @@ function byBbox(args, res) { // eslint-disable-line no-unused-vars
...
+
+      cassandraConnector.executeQuery(tagsQuery, tagsParams)
+      .then(rows => {
+        const eventIds = makeSet(rows.map(row => row.eventids), eventId => eventId);
+
+        const eventsQuery = `
+        SELECT *
+        FROM fortis.events
+        WHERE pipelinekey = ?
+        AND eventid IN ?
```
erikschlegel, line 90: apparently the limit of items you can include in a C* IN clause is 65535. We should be fine as long as we cap the number of filtered `eventid` based on a `LIMIT` threshold from ^^^.

src/resolvers-cassandra/Messages/queries.js
```diff
@@ -130,21 +48,55 @@ function byBbox(args, res) { // eslint-disable-line no-unused-vars
...
+        AND fulltext LIKE ?
+        `.trim();
+
+        const eventsParams = [
+          toPipelineKey(args.sourceFilter),
+          eventIds,
+          `%${args.fulltextTerm}%`
```
erikschlegel: 👍

src/resolvers-cassandra/Edges/queries.js
c-w: Done in 97a935a.

erikschlegel (approved):
Looks great. Sorry for the delay. Please opine to the comments below.

src/resolvers-cassandra/Edges/queries.js
c-w: Done in 4b4d59b.

src/resolvers-cassandra/Messages/queries.js
c-w: Done in 6bdb30d.

src/resolvers-cassandra/Edges/queries.js
```diff
@@ -70,45 +41,186 @@ function locations(args, res) { // eslint-disable-line no-unused-vars
...
+    const { period, periodType } = parseTimespan(args.timespan);
+
+    const query = `
+    SELECT placename, placeid, mentioncount
+    FROM fortis.popularplaces
+    WHERE period = ?
+    AND periodtype = ?
+    AND pipelinekey = ?
+    AND externalsourceid = ?
+    AND conjunctiontopic1 = ?
```
Smarker, line 87: I know this was a change for https://github.com/CatalystCode/project-fortis-pipeline/commit/f869c200e35e6d68ae920f10f0bed8f70cfacfbf, but `conjunctiontopic1` doesn't seem so clear to me just with the 1 appended. I see that it is derived from the first parameter `mainEdge` in `toConjunctionTopics`, so maybe there could be a better name for this? But then doing so conjunctiontopic2 and conjunctiontopic3 would have to be changed as well to reflect a change in name of conjunctiontopic1.

src/resolvers-cassandra/Messages/queries.js
c-w: Done in ab36b4c.

src/resolvers-cassandra/Edges/queries.js
c-w: This is one for @erikschlegel. I don't particularly care about the naming here because we have quite a few things in the database that require documentation already.

PR #89, opened 2017-08-07
Memoize locations for bounding box

c-w:
@erikschlegel mentioned that it would be good to get rid of the feature service call to fetch the locations inside of a bounding box. This is non-trivial to do by changing the schema of the eventplaces table.

```cql
CREATE TABLE eventplaces(
    eventids set<text>,
    conjunctiontopic1 text,
    conjunctiontopic2 text,
    conjunctiontopic3 text,
    placeid text,
    eventtime timestamp,
    pipelinekey text,
    externalsourceid text,
    PRIMARY KEY ((placeid), conjunctiontopic1, conjunctiontopic2, conjunctiontopic3, pipelinekey, externalsourceid, eventtime)
);
```

Currently the table is keyed off of placeid which we get back from the feature service after querying for all the places in the bounding box. If we change the table to include the place x/y coordinates, we'd need to include the x/y in the key to keep the uniqueness of every row. However, we can't use primary key columns in a range query:

pasted image at 2017_08_07 10_15 am

Simply caching the bounding box place results should give us quite a lot of the value already as the bounding box calls should always be for the same value as we don't expect the bounding box to be changed once the site is set up.

### DIFF
--- a/src/resolvers-cassandra/Edges/queries.js
+++ b/src/resolvers-cassandra/Edges/queries.js
@@ -3,7 +3,7 @@
 const Promise = require('promise');
 const cassandraConnector = require('../../clients/cassandra/CassandraConnector');
 const featureServiceClient = require('../../clients/locations/FeatureServiceClient');
-const { tilesForBbox, parseTimespan, parseFromToDate, withRunTime, toPipelineKey, toConjunctionTopics } = require('../shared');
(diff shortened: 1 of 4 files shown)


jcjimenez (approved):
LGTM

Smarker (approved):
LGTM

c-w:
@jcjimenez made a good point that we can just generate this list once and insert it into Cassandra. Will follow up with that approach.

PR #91, opened 2017-08-07
Pre-compute all places inside of the site geofence

c-w:
This is a follow-up to the abandoned pull request https://github.com/CatalystCode/project-fortis-services/pull/89

At site creation time, we do a single call to fetch all the place ids inside of the geofence defined for the Fortis deployment. Later, we can then just read back these place detail from Cassandra instead of having
to hit the feature service over and over again. This is desirable over the approach taken in the linked pull request as we only have to hit the feature service once ever (instead of once per user session) and because
we reduce the amount of data that we have to cache in memory on the GraphQL server (there may be many places in a geofence).

The requisite schema change is in https://github.com/CatalystCode/project-fortis-pipeline/pull/96


erikschlegel (approved):
LGTM

PR #92, opened 2017-08-08
Remove event_time and eventids

c-w:
@erikschlegel mentioned that we should avoid the eventids collection as the collections hold at most 65k entries and we may have more events for a particular point-in-time and topic/place combination.

Doing the change on the query side first, will do the change in the schema repository once we've agreed on an approach.


Smarker (approved):
LGTM

jcjimenez (approved):
LGTM

PR #93, opened 2017-08-09
Restore byLocations query

c-w:
This is the service-layer change for this frontend change:
https://github.com/CatalystCode/project-fortis-interfaces/pull/21

### DIFF
--- a/src/resolvers-cassandra/Messages/queries.js
+++ b/src/resolvers-cassandra/Messages/queries.js
@@ -31,12 +31,66 @@ function eventToFeature(row) {
 }
 
 /**
- * @param {site: string, originalSource: string, coordinates: number[], filteredEdges: string[], langCode: string, limit: number, offset: number, fromDate: string, toDate: string, sourceFilter: string[], fulltextTerm: string} args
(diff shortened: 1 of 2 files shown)


src/resolvers-cassandra/Messages/queries.js
```diff
@@ -31,12 +31,66 @@ function eventToFeature(row) {
...
+    .then(rows => {
+      const eventIds = makeSet(rows, row => row.eventid);
+
+      const eventsQuery = `
+      SELECT *
+      FROM fortis.events
+      WHERE pipelinekey = ?
+      AND eventid IN ?
+      AND fulltext LIKE ?
+      LIMIT ?
```
jcjimenez, line 76: Does the `LIMIT` clause here and the absence of explicit cluster ordering in the definition of the `events` table mean this would yield a random collection of values?

jcjimenez (approved):
LGTM with questions

src/resolvers-cassandra/Messages/queries.js
c-w: The order would be defined by the clustering columns. https://stackoverflow.com/a/21949250/3817588

erikschlegel (approved):
LGTM

c-w:
Approach is made impossible by conversation around https://github.com/CatalystCode/project-fortis-services/pull/94

PR #96, opened 2017-08-15
Fix cassandra auth

c-w:
The cluster cassandra has password authentication enabled so the client also needs to support this. Without this configuration, we get errors like this:

image


Smarker (approved):
LGTM

PR #98, opened 2017-08-15
Block server start until Cassandra is initialized

c-w:
The first request to Cassandra is very slow (on the order of 20 seconds
during my testing), I reckon because the connection and pool are being
built up. Subsequent requests are much faster (sub-second).

We don't want our users to have a problem with hitting the slow first
request, so we should warm up the Cassandra connection before we let the
server start. In this way, when we re-deploy the GraphQL server, we'll
simply continue running the instances of the old code in the load
balancer until the new server is fully operational and has warmed up the
Cassandra connection.


Smarker (approved):
LGTM

PR #99, opened 2017-08-15
Fix edges queries for cluster cassandra

c-w:
Depends on this schema change: https://github.com/CatalystCode/project-fortis-pipeline/pull/110

### DIFF
--- a/src/clients/locations/FeatureServiceClient.js
+++ b/src/clients/locations/FeatureServiceClient.js
@@ -79,6 +79,7 @@ function fetchByPoint(point, extraFields) {
  */
 function fetchByName(name, extraFields) {
   const names = name.constructor === Array ? name : [name];
+  if (!names.length) return Promise.reject('No names specified to fetch; no places match query');
   return callFeatureService(formatNameUri(names, extraFields));
 }
 
@@ -89,6 +90,7 @@ function fetchByName(name, extraFields) {
  */
 function fetchById(id, extraFields) {
   const ids = id.constructor === Array ? id : [id];
+  if (!ids.length) return Promise.reject('No ids specified to fetch; no places match query');
   return callFeatureService(formatIdsUri(ids, extraFields));
 }
 
--- a/src/clients/locations/FeatureServiceClient.js
+++ b/src/clients/locations/FeatureServiceClient.js
@@ -6,27 +6,27 @@ const trackDependency = require('../appinsights/AppInsightsClient').trackDepende
 
 const apiUrlBase = process.env.FORTIS_FEATURE_SERVICE_HOST;
 
-function formatIdsUri(ids, include) {
+function formatIdsUri(ids, extraFields) {
   let uri = `${apiUrlBase}/features/id/${ids.map(encodeURIComponent).join(',')}`;
-  if (include) uri += `?include=${include}`;
+  if (extraFields) uri += `?include=${extraFields}`;
   return uri;
 }
 
-function formatBboxUri(north, west, south, east, include) {
+function formatBboxUri(north, west, south, east, extraFields) {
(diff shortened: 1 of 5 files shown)


src/clients/locations/FeatureServiceClient.js
```diff
@@ -6,20 +6,28 @@ const trackDependency = require('../appinsights/AppInsightsClient').trackDepende
 
 const apiUrlBase = process.env.FORTIS_FEATURE_SERVICE_HOST;
 
-function formatIdsUri(ids) {
-  return `${apiUrlBase}/features/id/${ids.map(encodeURIComponent).join(',')}?include=bbox`;
+function formatIdsUri(ids, include) {
+  let uri = `${apiUrlBase}/features/id/${ids.map(encodeURIComponent).join(',')}`;
```
Smarker: What about using `const` for `uri`? (there's a couple of other instances where there's `let uri`)

src/clients/locations/FeatureServiceClient.js
```diff
@@ -6,20 +6,28 @@ const trackDependency = require('../appinsights/AppInsightsClient').trackDepende
...
+  if (include) uri += `?include=${include}`;
```
Smarker: Not really sure what `include` is from the name.

src/clients/locations/FeatureServiceClient.js
```diff
@@ -6,20 +6,28 @@ const trackDependency = require('../appinsights/AppInsightsClient').trackDepende
 
 const apiUrlBase = process.env.FORTIS_FEATURE_SERVICE_HOST;
 
-function formatIdsUri(ids) {
-  return `${apiUrlBase}/features/id/${ids.map(encodeURIComponent).join(',')}?include=bbox`;
+function formatIdsUri(ids, include) {
```
Smarker: Does `include` have to be a parameter passed in to all of these methods? Could it be a member of the `FeatureServiceClient`?

src/resolvers-cassandra/Edges/queries.js
```diff
@@ -76,10 +78,10 @@ function locations(args, res) { // eslint-disable-line no-unused-vars
  */
 function popularLocations(args, res) { // eslint-disable-line no-unused-vars
   return new Promise((resolve, reject) => {
-    const { period, periodType } = parseTimespan(args.timespan);
+    const { period, periodType, fromDate, toDate } = parseTimespan(args.timespan);
 
     const query = `
-    SELECT placename, placeid, mentioncount
+    SELECT placeid, mentioncount, centroidlat, centroidlon
```
erikschlegel: This query should factor in the bbox coordinates.

Smarker (approved):
LGTM - with comments

src/resolvers-cassandra/Edges/queries.js
c-w: Added new feature in 4db00c6. Will also require client-side change since bbox was not considered in Fortis-v1.

src/clients/locations/FeatureServiceClient.js
c-w: FeatureServiceClient is not an object, these are all just functions. What we want to include depends on the context of the request (e.g. some callers need the bbox, others don't), so this really should be configurable on a per-call level.

src/clients/locations/FeatureServiceClient.js
c-w: The uri is being mutated just below so it can't be const :)

src/clients/locations/FeatureServiceClient.js
c-w: Renamed in  436c114.

PR #100, opened 2017-08-15
Fix messages queries with cluster cassandra

c-w:
Pull request relies on this schema change: https://github.com/CatalystCode/project-fortis-pipeline/pull/111

### DIFF
--- a/src/resolvers-cassandra/Messages/queries.js
+++ b/src/resolvers-cassandra/Messages/queries.js
@@ -3,7 +3,7 @@
 const Promise = require('promise');
 const translatorService = require('../../clients/translator/MsftTranslator');
 const cassandraConnector = require('../../clients/cassandra/CassandraConnector');
-const { parseFromToDate, withRunTime, toPipelineKey, toConjunctionTopics, limitForInClause } = require('../shared');
+const { parseFromToDate, parseLimit, withRunTime, toPipelineKey, toConjunctionTopics, limitForInClause } = require('../shared');
 const { makeSet } = require('../../utils/collections');
 const trackEvent = require('../../clients/appinsights/AppInsightsClient').trackEvent;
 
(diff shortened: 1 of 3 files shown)


src/resolvers-cassandra/Messages/queries.js
```diff
@@ -51,8 +93,8 @@ function byLocation(args, res) { // eslint-disable-line no-unused-vars
     AND externalsourceid = ?
     AND centroidlat = ?
     AND centroidlon = ?
-    AND eventtime >= ?
-    AND eventtime <= ?
+    AND eventime >= ?
```
Smarker: We're not using `eventtime` anymore?

src/resolvers-cassandra/Messages/queries.js
```diff
@@ -264,8 +242,16 @@ function event(args, res) { // eslint-disable-line no-unused-vars
...
   return new Promise((resolve, reject) => {
     translatorService.translate(args.sentence, args.fromLanguage, args.toLanguage)
-      .then(result => resolve({ translatedSentence: result.translatedSentence, originalSentence: args.sentence }))
-      .catch(reject);
+    .then(result => {
+      const translatedSentence = result.translatedSentence;
+      const originalSentence = args.sentence;
+
+      resolve({
+        translatedSentence,
```
Smarker, line 250: You can have an object without properties?

c-w: This is one of the ES6 niceties:

```js
const foo = 1;
{foo}
```

Is the same as:

```js
const foo = 1;
{foo: foo}
```

Only less verbose :)

src/resolvers-cassandra/Messages/queries.js
c-w: Good catch, I made this change to enable testing against the cluster but it should be reverted.

c-w: Fixed in f736292.

src/resolvers-cassandra/Messages/queries.js
Smarker: Ok good to know!

Smarker (approved):
LGTM

PR #103, opened 2017-08-17
Rename insertion_time to insertiontime

Smarker:
This change is needed because `insertion_time` was changed to `insertiontime` in `cassandra-setup.cql`
see: https://github.com/CatalystCode/project-fortis-pipeline/blob/master/ops/storage-ddls/cassandra-setup.cql#L81


jcjimenez (approved):
LGTM

PR #104, opened 2017-08-17
Implement createStream and edit SettingsSchema

Smarker:
`removeStream` will be done in a different PR.


Smarker:
image

Smarker:
Just realized I was missing the modify part.

PR #105, opened 2017-08-17
Implement createStream and edit SettingsSchema

Smarker:
`removeStream` will be done in a separate PR (https://github.com/CatalystCode/project-fortis-services/pull/107)

### DIFF
--- a/src/resolvers-cassandra/Settings/mutations.js
+++ b/src/resolvers-cassandra/Settings/mutations.js
@@ -141,29 +141,38 @@ function paramEntryToMap(paramEntry) {
  */
 function createStream(args, res) { // eslint-disable-line no-unused-vars
   return new Promise((resolve, reject) => {
+    const pipelineKey = args && args.input && args.input.pipelineKey;
+    if (!pipelineKey && !pipelineKey.length) return reject('No pipelineKey specified.');
+    
     const params = paramEntryToMap(args.input.params);
     cassandraConnector.executeBatchMutations([{
-      query: `INSERT INTO fortis.streams (
-      streamid, 
-      pipelinekey,
-      pipelinelabel,
-      pipelineicon,
-      streamfactory,
-      params
-      ) VALUES (?, ?, ?, ?, ?, ?)`,
-      params: [
-        uuid(), 
-        args.input.pipelineKey,
-        args.input.pipelineLabel,
-        args.input.pipelineIcon,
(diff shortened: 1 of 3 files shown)


Smarker:
image

src/schemas/SettingsSchema.js
```diff
@@ -13,6 +13,7 @@ module.exports = graphql.buildSchema(`
   type Mutation {
     createSite(input: SiteDefinition!): Site
     removeSite(input: SiteDefinition!): Site
+    createStream(input: StreamDefinition!): Stream
```
c-w, line 16: Could this return a StreamDefinition so that you don't need to have both types around?

src/resolvers-cassandra/Settings/mutations.js
```diff
@@ -131,6 +131,58 @@ function removeSite(args, res) { // eslint-disable-line no-unused-vars
...
+}
+
+/**
+ * @param {{input: {pipelineKey: string, pipelineLabel: string, pipelineIcon: string, streamFactory: string, params: Array<{String: String}>}}} args
+ * @returns {Promise}
+ */
+function createStream(args, res) { // eslint-disable-line no-unused-vars
+  return new Promise((resolve, reject) => {
+    const pipelineKey = args && args.input && args.input.pipelineKey;
+    if (!pipelineKey && !pipelineKey.length) return reject('No pipelineKey specified.');
```
c-w: Should the check be an `||`?

c-w (approved):

src/resolvers-cassandra/Settings/mutations.js
Smarker: yes done in [e072637](https://github.com/CatalystCode/project-fortis-services/pull/105/commits/e072637463445c28f83a0a8117704734fa77061a)

src/schemas/SettingsSchema.js
Smarker: I had to include a separate `Stream` type. If I returned a `StreamDefinition`, the `StreamDefinition` is an `input` type and the `params` fields holds an input type `ParamsEntry`, which results in an expected output type error: 
```
Error: Expected Output type.
    at invariant (C:\Users\stmarker\Documents\project-fortis\project-fortis-services\node_modules\graphql\jsutils\invariant.js:19
:11)
```

PR #109, opened 2017-08-17
Fix bug with undefined params field

Smarker:
Editing graphql schema to make some `StreamDefinition` fields required.

### DIFF
--- a/src/resolvers-cassandra/Settings/mutations.js
+++ b/src/resolvers-cassandra/Settings/mutations.js
@@ -143,11 +143,11 @@ function createStream(args, res) { // eslint-disable-line no-unused-vars
   return new Promise((resolve, reject) => {
     const pipelineKey = args && args.input && args.input.pipelineKey;
     if (!pipelineKey || !pipelineKey.length) return reject('No pipelineKey specified.');
-    
-    let params = {};
+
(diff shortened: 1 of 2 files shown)


c-w (commented):
Would it make sense to create a new connector without any params? Seems unlikely. Should the input params entries list be require?

Smarker:
yea I think it should be required. I'll update the schema accordingly

src/resolvers-cassandra/Settings/mutations.js
```diff
@@ -143,8 +143,11 @@ function createStream(args, res) { // eslint-disable-line no-unused-vars
   return new Promise((resolve, reject) => {
     const pipelineKey = args && args.input && args.input.pipelineKey;
     if (!pipelineKey || !pipelineKey.length) return reject('No pipelineKey specified.');
-    
-    const params = paramEntryToMap(args.input.params);
+
+    const paramEntry = args && args.input && args.input.params;
+    if (!paramEntry || !paramEntry.length || paramEntry.length == 0) return reject('No params specified.');
```
c-w: If you make the param required, no need for this check as graphql will reject a request that doesn't contain the param.

c-w (approved):

PR #111, opened 2017-08-18
Augment return value with coordinates if possible

c-w:


src/resolvers-cassandra/Messages/queries.js
```diff
@@ -2,19 +2,22 @@
...
  * @typedef {type: string, coordinates: number[][], properties: {edges: string[], messageid: string, createdtime: string, sentiment: number, title: string, originalSources: string[], sentence: string, language: string, source: string, properties: {retweetCount: number, fatalaties: number, userConnecionCount: number, actor1: string, actor2: string, actor1Type: string, actor2Type: string, incidentType: string, allyActor1: string, allyActor2: string, title: string, link: string, originalSources: string[]}, fullText: string}} Feature
  */
 
-function eventToFeature(row) {
+function eventToFeature(row, placeIdToCentroid) {
   return {
     type: row.pipelinekey,
-    coordinates: [],
+    coordinates: row.placeids && row.placeids.length > 0
+      ? row.placeids.map(placeid => placeIdToCentroid[placeid]).filter(centroid => !!centroid)
```
erikschlegel, line 19: I believe we already have the centroid coordinates available in the events table https://github.com/CatalystCode/project-fortis-pipeline/blob/master/ops/storage-ddls/cassandra-setup.cql#L71 . Is there another reason why we need to make a trip to the feature service for each event?

PR #113, opened 2017-08-21
Fixed admin setting integration issues

erikschlegel:


src/resolvers-cassandra/Settings/mutations.js
```diff
@@ -468,6 +518,7 @@ module.exports = {
   createSite: trackEvent(createSite, 'createSite'),
   removeSite: trackEvent(removeSite, 'removeSite'),
   createStream: trackEvent(createStream, 'createStream'),
+  editSite: trackEvent(withRunTime(editSite), 'editSite'),
```
c-w, line 521: Is it worth having a separate editSite when it can be implemented client-side by first issuing rmeoveSite and then createSite?

src/resolvers-cassandra/Settings/queries.js
```diff
@@ -35,12 +35,14 @@ function sites(args, res) { // eslint-disable-line no-unused-vars
     if (!siteId) {
       return reject('No site id to fetch specified');
     }
+    
+    console.log('called')
```
c-w: Please remove debugging code as it's causing the linter to fail.

src/resolvers-cassandra/Settings/mutations.js
Smarker: Is it worth adding `withRunTime` to track an operation to a single site or stream? If so, I'll need to add that to `createSite`, `removeSite`, `createStream`, `removeStream`

erikschlegel: @c-w we shouldnt have to remove a site to edit a single property. If we followed your suggestion and the `createSite` fails after `removeSite` succeeds then the entire pipeline is down due to C* lack of transaction support.

jcjimenez (approved):
LGTM

PR #114, opened 2017-08-21
Use default appinsights environment variable name

c-w:
As per conversation on Slack: we decide to use a single appinsights account for all services so might as well use the same environment variable name everywhere.


erikschlegel (approved):
LGTM

README.md
```diff
@@ -18,7 +18,7 @@ TRANSLATION_SERVICE_ACCOUNT_KEY="..."
 FACEBOOK_AUTH_TOKEN="..."
 
 FORTIS_FEATURE_SERVICE_HOST="..."
-FORTIS_SERVICES_APPINSIGHTS_KEY="..."
+APPINSIGHTS_INSTRUMENTATIONKEY="..."
```
jcjimenez, line 21: I noticed that in `project-fortis-spark` the environment variable is `APPLICATION_INSIGHTS_IKEY`. Do we have a PR for that project as well? Or should we just make this project use that key?

jcjimenez (approved):
LGTM with question

README.md
c-w: This is amazing. The AppInsights SDKs are inconsistent between each other. Java uses APPLICATION_INSIGHTS_IKEY and all the others use APPINSIGHTS_INSTRUMENTATIONKEY. The latter is recommended, so I think we can stick with that.

PR #115, opened 2017-08-22
A combination of fixes and added features to support fortis events in the newsfeed

erikschlegel:


src/clients/cassandra/CassandraConnector.js
```diff
@@ -93,6 +93,30 @@ function executeQuery(query, params, options) {
...
+ * @param {int} fetchSize
+ * @param {string} pageState
+ * @returns {Promise.<object[]>}
+ */
+function executeQueryWithPageState(query, params, pageState, fetchSize) {
+  return new Promise((resolve, reject) => {
+      const DEFAULT_FETCH = 15;
+
+      let options = {
+        fetchSize: fetchSize || DEFAULT_FETCH
```
c-w: This may be better as `fetchSize > 0 ? fetchSize : DEFAULT_FETCH` since it'll handle the degenerate case where someone calls this with a negative fetch size.

src/clients/cassandra/CassandraConnector.js
```diff
@@ -93,6 +93,30 @@ function executeQuery(query, params, options) {
...
+      };
+
+      if(pageState)
```
c-w: Nit: we use space before opening parens and curly braces around statements, so:

```
if (pageState) {
  options.pageState = pageState
}
```

Added question: would this be different from always setting the pageState? If it's undefined then the downstream library should just be able to handle that just as if it wasn't set:

```
options = {
  pageState: pageState || undefined
}
```

src/clients/cassandra/CassandraConnector.js
```diff
@@ -93,6 +93,30 @@ function executeQuery(query, params, options) {
...
+        options.pageState = pageState
+
+      client.execute(query, params, options)
+            .then(result => resolve(Object.assign({}, {pageState: result.pageState, rows: result.rows})))
```
c-w: Nit: elsewhere aligns the promise continuation methods with the first line, so:

```
client.execute(...)
.then(...)
.catch(...)
```

src/clients/cassandra/CassandraConnector.js
```diff
@@ -93,6 +93,30 @@ function executeQuery(query, params, options) {
...
+            .catch(exception =>  reject(exception))
```
c-w: Nit: this can be simplified to `.catch(reject)`

src/resolvers-cassandra/Messages/queries.js
```diff
@@ -12,26 +12,33 @@ const trackEvent = require('../../clients/appinsights/AppInsightsClient').trackE
  */
 
 function eventToFeature(row) {
+  const FeatureType = 'MultiPoint';
+
   return {
-    type: row.pipelinekey,
-    coordinates: [],
+    type: FeatureType,
+    coordinates: row.computedfeatures.places.map(place=>[place.centroidlon, place.centroidlat]),
```
c-w, line 19: This is also implemented here: https://github.com/CatalystCode/project-fortis-services/pull/111

src/resolvers-cassandra/Messages/queries.js
```diff
@@ -12,26 +12,33 @@ const trackEvent = require('../../clients/appinsights/AppInsightsClient').trackE
  */
 
 function eventToFeature(row) {
+  const FeatureType = 'MultiPoint';
```
c-w, line 15: General question: what's the point of having this if there's only one acceptable value? The schema defines this like so:

```
  enum FeatureType {
    MultiPoint
  }
```

src/resolvers-cassandra/Messages/queries.js
```diff
@@ -12,26 +12,33 @@ const trackEvent = require('../../clients/appinsights/AppInsightsClient').trackE
...
+      sentiment: row.computedfeatures && row.computedfeatures.sentiment ? row.computedfeatures.sentiment.neg_avg : -1,
       title: row.title,
+      fullText: row.body,
+      //snippet: todo: bind field once it's available in the pipeline
       originalSources: row.externalsourceid,
       language: row.eventlangcode,
-      source: row.sourceurl,
-      fullText: row.messagebody
+      source: row.pipelinekey,
+      fullText: row.body,
```
c-w: Field is duplicate with above.

src/resolvers-cassandra/shared.js
```diff
@@ -87,7 +87,9 @@ function parseFromToDate(fromDate, toDate) { // eslint-disable-line no-unused-va
 }
 
 function parseLimit(limit) {
-  return limit > 0 ? limit : 15;
+  const DEFAULT_LIMIT = 15;
+
+  return limit && limit > 0 ? limit : DEFAULT_LIMIT;
```
c-w: This can just be `limit > 0` since undefined and null are non-positive:

image

src/resolvers-cassandra/Messages/queries.js
c-w: NB: wouldn't these be placeids from the events table? https://github.com/CatalystCode/project-fortis-pipeline/blob/32e8c1795a787db6ce4ae181e746d4d754f4ae19/ops/storage-ddls/cassandra-setup.cql#L249

src/clients/cassandra/CassandraConnector.js
```diff
@@ -93,6 +93,30 @@ function executeQuery(query, params, options) {
...
+ * @param {int} fetchSize
+ * @param {string} pageState
+ * @returns {Promise.<object[]>}
+ */
+function executeQueryWithPageState(query, params, pageState, fetchSize) {
+  return new Promise((resolve, reject) => {
+    const DEFAULT_FETCH = 15;
+
+    let options = {
+      fetchSize: fetchSize || DEFAULT_FETCH
```
c-w: This may be better as fetchSize > 0 ? fetchSize : DEFAULT_FETCH since it'll handle the degenerate case where someone calls this with a negative fetch size.

src/clients/cassandra/CassandraConnector.js
```diff
@@ -93,6 +93,30 @@ function executeQuery(query, params, options) {
...
+    };
+
+    if(pageState)
```
c-w: Nit: we use space before opening parens and curly braces around statements, so:

```
if (pageState) {
  options.pageState = pageState
}
```

Added question: would this be different from always setting the pageState? If it's undefined then the downstream library should just be able to handle that just as if it wasn't set:

```
options = {
  pageState: pageState || undefined
}
```

src/clients/cassandra/CassandraConnector.js
```diff
@@ -93,6 +93,30 @@ function executeQuery(query, params, options) {
...
+      options.pageState = pageState;
+
+    client.execute(query, params, options)
+          .then(result => resolve(Object.assign({}, {pageState: result.pageState, rows: result.rows})))
```
c-w: Nit: elsewhere aligns the promise continuation methods with the first line, so:

```
client.execute(...)
.then(...)
.catch(...)
```

src/clients/cassandra/CassandraConnector.js
```diff
@@ -93,6 +93,30 @@ function executeQuery(query, params, options) {
...
+          .catch(exception =>  reject(exception));
```
c-w: Nit: this can be simplified to `.catch(reject)`

src/resolvers-cassandra/Messages/queries.js
```diff
@@ -119,46 +120,46 @@ function byLocation(args, res) { // eslint-disable-line no-unused-vars
...
 /**
- * @param {site: string, originalSource: string, bbox: number[], mainTerm: string, filteredEdges: string[], langCode: string, limit: number, offset: number, fromDate: string, toDate: string, sourceFilter: string[], fulltextTerm: string} args
+ * @param {externalsourceid: string, bbox: number[], conjunctivetopics: string[], limit: number, pageState: number, fromDate: string, toDate: string, pipelinekeys: string[], fulltextTerm: string} args
  * @returns {Promise.<{runTime: string, type: string, bbox: number[], features: Feature[]}>}
- */
+ **/
 function byBbox(args, res) { // eslint-disable-line no-unused-vars
   return new Promise((resolve, reject) => {
     if (!args.bbox || args.bbox.length !== 4) return reject('Invalid bbox specified');
+    if(!args.conjunctivetopics || args.conjunctivetopics.length === 0) return reject('Empty conjunctive topic list specified');
```
c-w: Nit: space before parens.

src/resolvers-cassandra/Messages/queries.js
```diff
@@ -119,46 +120,46 @@ function byLocation(args, res) { // eslint-disable-line no-unused-vars
...
+      ...fromTopicListToConjunctionTopics(args.conjunctivetopics),
+      args.toDate,
+      north,
+      east,
+      args.fromDate,
+      south,
+      west
+    ];
+
+    if(args.externalsourceid){
```
c-w: Nit: space before parens and squiggly bracket.

src/resolvers-cassandra/Messages/queries.js
```diff
@@ -119,46 +120,46 @@ function byLocation(args, res) { // eslint-disable-line no-unused-vars
...
     AND conjunctiontopic2 = ?
     AND conjunctiontopic3 = ?
-    AND pipelinekey = ?
-    AND externalsourceid = ?
-    AND (centroidlat, centroidlon, eventtime) <= (?, ?, ?)
-    AND (centroidlat, centroidlon, eventtime) >= (?, ?, ?)
-    LIMIT ?
+    AND (eventtime, centroidlat, centroidlon) <= (?, ?, ?)
+    AND (eventtime, centroidlat, centroidlon) >= (?, ?, ?)
+    ${args.externalsourceid ? ' AND externalsourceid = ?' : ''}
```
c-w: I'm surprised that this query works. Could you please explain how that is possible? From my understanding, all equal columns have to come before all non-equal columns. https://stackoverflow.com/a/36752715/3817588

src/resolvers-cassandra/Messages/queries.js
```diff
@@ -119,46 +120,46 @@ function byLocation(args, res) { // eslint-disable-line no-unused-vars
...
+    ${args.pipelinekeys ? ` AND pipelinekey IN('${args.pipelinekeys.join('\',\'')}')` : ''}
```
c-w: It would be safer to insert placeholders (`?`) here and add the actual pipelinekey values via the params array.

src/resolvers-cassandra/shared.js
```diff
@@ -44,6 +44,21 @@ function toPipelineKey(sourceFilter) {
   return sourceFilter[0];
 }
 
+function fromTopicListToConjunctionTopics(topicTerms){
```
c-w: Nit: space before squiggly bracket.

src/clients/cassandra/CassandraConnector.js
```diff
@@ -93,6 +93,33 @@ function executeQuery(query, params, options) {
...
+
+    let options = {
+      fetchSize: fetchSize > 0 ? fetchSize : DEFAULT_FETCH
+    };
+
+    if(pageState) {
+      options.pageState = pageState;
+    }
+
+    console.log(query);
```
c-w: Please remove debug code as this will cause the lint to fail.

src/clients/cassandra/CassandraConnector.js
```diff
@@ -93,6 +93,33 @@ function executeQuery(query, params, options) {
...
+
+    client.execute(query, params, options)
+    .then(result => resolve(Object.assign({}, {pageState: result.pageState, rows: result.rows})))
```
c-w: Why the need for Object.assign instead of just declaring a new object? Isn't the following simpler?

```js
.then(result => {
  const pageState = result.pageState;
  const rows = result.rows;

  resolve({
    pageState,
    rows
  })
})
```

If you want it more concise, it's also possible to inline the variables:

```js
.then(result => ({pageState: result.pageState, rows: result.rows}))
```

src/clients/cassandra/CassandraConnector.js
```diff
@@ -93,6 +93,33 @@ function executeQuery(query, params, options) {
...
+    .catch(exception =>  reject(exception));
```
c-w: This is equivalent to `.catch(reject)` which we use elsewhere.

src/clients/cassandra/CassandraConnector.js
```diff
@@ -93,6 +93,33 @@ function executeQuery(query, params, options) {
...
+ */
+function executeQueryWithPageState(query, params, pageState, fetchSize) {
+  return new Promise((resolve, reject) => {
+    const DEFAULT_FETCH = 15;
+
+    let options = {
+      fetchSize: fetchSize > 0 ? fetchSize : DEFAULT_FETCH
+    };
+
+    if(pageState) {
```
c-w: Nit: space before parens:

```js
if (pageState) {
```

src/resolvers-cassandra/Messages/queries.js
```diff
@@ -54,18 +58,15 @@ function queryEventsTable(rowsWithEventIds, args) {
...
-
     cassandraConnector.executeQuery(eventsQuery, eventsParams)
     .then(rows => {
       const features = rows.map(eventToFeature);
+      const pageState = eventIdResponse.pageState;
 
       resolve({
-        features
+        type: 'FeatureCollection',
+        features, pageState,
```
c-w: Nit: One property per line is easier to read.

```
resolve({
  type: 'FeatureCollection',
  features,
  pageState,
  bbox: args.bbox
})
```

Nit: it seems a bit arbitrary why some of the properties are defined inline and some are defined as const variables above. Elsewhere in the project this is solved by just always declaring the variables that get resolved as const, so:

```js
const features = rows.map(eventToFeature);
const pageState = eventIdResponse.pageState;
const bbox = args.bbox;
const type = 'FeatureCollection';

resolve({
  type,
  features,
  pageState,
  bbox
});
```

Don't particularly care either way, just wanted to point out the inconsistency.

src/resolvers-cassandra/Messages/queries.js
```diff
@@ -119,46 +120,47 @@ function byLocation(args, res) { // eslint-disable-line no-unused-vars
...
 /**
- * @param {site: string, originalSource: string, bbox: number[], mainTerm: string, filteredEdges: string[], langCode: string, limit: number, offset: number, fromDate: string, toDate: string, sourceFilter: string[], fulltextTerm: string} args
+ * @param {externalsourceid: string, bbox: number[], conjunctivetopics: string[], limit: number, pageState: number, fromDate: string, toDate: string, pipelinekeys: string[], fulltextTerm: string} args
  * @returns {Promise.<{runTime: string, type: string, bbox: number[], features: Feature[]}>}
- */
+ **/
 function byBbox(args, res) { // eslint-disable-line no-unused-vars
   return new Promise((resolve, reject) => {
     if (!args.bbox || args.bbox.length !== 4) return reject('Invalid bbox specified');
+    if (!args.conjunctivetopics || args.conjunctivetopics.length === 0) return reject('Empty conjunctive topic list specified');
```
c-w: This is safer as `!args.conjunctivetopics.length` which catches the case where conjunctivetopics is empty but also when it's not an array.

src/resolvers-cassandra/Messages/queries.js
```diff
@@ -119,46 +120,47 @@ function byLocation(args, res) { // eslint-disable-line no-unused-vars
...
     AND conjunctiontopic2 = ?
     AND conjunctiontopic3 = ?
-    AND pipelinekey = ?
-    AND externalsourceid = ?
-    AND (centroidlat, centroidlon, eventtime) <= (?, ?, ?)
-    AND (centroidlat, centroidlon, eventtime) >= (?, ?, ?)
-    LIMIT ?
+    AND (eventtime, centroidlat, centroidlon) <= (?, ?, ?)
+    AND (eventtime, centroidlat, centroidlon) >= (?, ?, ?)
+    ${args.pipelinekeys ? ` AND pipelinekey IN ?` : ''}
```
c-w: Do we need to add something to the tagsParams array if `args.pipelinekeys` is truthy (like we do for the externalsourceid on line 145-148 above)?

src/resolvers-cassandra/Messages/queries.js
```diff
@@ -119,46 +116,47 @@ function byLocation(args, res) { // eslint-disable-line no-unused-vars
...
+    let tableName = 'eventplaces';
+    let tagsParams = [
+      ...fromTopicListToConjunctionTopics(args.conjunctivetopics),
+      args.toDate,
+      north,
+      east,
+      args.fromDate,
+      south,
+      west,
+      args.pipelinekeys
```
c-w, line 138: Shouldn't this only be included if the `args.pipelinekeys` is truthy? Otherwise we'll have the extra undefined param in here but nothing to bind it to in the query below.

Example to make it clearer:
1) Assume args.pipelinekey is null
2) The tagsParams array will be [..., null] since args.pipelinekey is null.
3) The tagsQuery will not include the `AND pipelinekey in ?` since it's guarded by the if statement.
4) Now we have N params to bind in tagsParams and N-1 placeholders (`?`) in the query so there should be an error.

To fix it, I would suggest this:

```js
const tagsParams = [
  ...fromTopicListToConjunctionTopics(args.conjunctivetopics),
  args.toDate,
  north,
  east,
  args.fromDate,
  south,
  west,
];

if (args.pipelinekey) {
  tagsParams.push(args.pipelinekey);
}
```

src/schemas/MessageSchema.js
```diff
@@ -72,30 +79,16 @@ module.exports = graphql.buildSchema(`
...
     messageid: ID,
-    createdtime: String,
+    eventtime: String,
+    sourceeventid: String,
     sentiment: Float,
+    entities: [String],
     title: String,
-    originalSources: [String],
-    sentence: String,
+    externalsourceid: String,
```
c-w, line 87: This will also require some changes in project-fortis-interfaces. Would be ideal to have them before merging this PR so that we can match up things and review related changes at the same time.

src/resolvers-cassandra/Messages/queries.js
Smarker: I'd look into this more since I think Clemens may have an implementation that can be reused.

Smarker (approved):
LGTM

c-w:
Also linking some comments that may have gotten lost in the flurry of changes:
- https://github.com/CatalystCode/project-fortis-services/pull/115#discussion_r134582177
- https://github.com/CatalystCode/project-fortis-services/pull/115#discussion_r134587781
- https://github.com/CatalystCode/project-fortis-services/pull/115#discussion_r134582019
- https://github.com/CatalystCode/project-fortis-services/pull/115#discussion_r134618808

c-w:
To reply to @Smarker's comment above https://github.com/CatalystCode/project-fortis-services/pull/115#discussion_r134641685, if we have all the data available, the way that this PR implements the population of the coordinates field is much better since it doesn't cause extra trips to the feature-service. However, not sure if we have all the requisite data available in the events table where we only have placeid (not centroids), in which case we'd need something like #111.

PR #116, opened 2017-08-23
Stream query

Smarker:
Added query for streams to fetch all streams.

Will edit `createStream` and `removeStream` to support change in stream type in settings schema in a separate PR.


c-w (commented):
Change looks good. Please also link the equivalent frontend changes so they can be reviewed in parallel before merging.

Smarker:
Equivalent front end changes that use streams query: https://github.com/CatalystCode/project-fortis-interfaces/pull/28

c-w (approved):

PR #117, opened 2017-08-23
Add siteType to EditableSiteSettings

Smarker:
* `siteType` is needed for `createSite`, so that when a site is first created, the site can have predefined topics according to its siteType

* I made `siteType` optional, since it shouldn't be required for `removeSite` or `editSite`.

### DIFF
--- a/src/schemas/SettingsSchema.js
+++ b/src/schemas/SettingsSchema.js
@@ -132,6 +132,7 @@ module.exports = graphql.buildSchema(`
   }
 
   input EditableSiteSettings {
+    siteType: String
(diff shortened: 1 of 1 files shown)


src/schemas/SettingsSchema.js
```diff
@@ -132,6 +132,7 @@ module.exports = graphql.buildSchema(`
   }
 
   input EditableSiteSettings {
+    siteType: String
```
c-w: Missing comma at the end.

c-w (approved):

PR #118, opened 2017-08-25
Refactoring Edge Schema / Migrate Admin Services from Edges Schema -> Settings Schema

erikschlegel:


src/resolvers-cassandra/Edges/queries.js
```diff
@@ -8,71 +8,6 @@ const { makeSet, makeMap, makeMultiMap } = require('../../utils/collections');
...
-    })
-    .catch(reject);
-  });
-}
-
-/**
- * @param {{site: string}} args
- * @returns {Promise.<{runTime: string, edges: Array<{name: string, coordinates: number[]}>}>}
- */
-function locations(args, res) { // eslint-disable-line no-unused-vars
```
c-w, line 41: We may need to keep this method around since
(1) it's still referenced in Edges/index.js on line 6 above
(2) it's used by the frontend in services.js/fetchEdges which is used by load_localities

c-w (approved):
Approving to unblock you, but please address this comment: https://github.com/CatalystCode/project-fortis-services/pull/118#discussion_r135293720

PR #126, opened 2017-09-01
Remove nesting of .then on promise.

Smarker:


jcjimenez (approved):
LGTM - trusting the assignment of mutations works!

Smarker:
image

I believe it works, since I still get new topics populated on `createSite` on my local cassandra.

PR #127, opened 2017-09-05
Integrated Cassandra Refactored Tables with Services

erikschlegel:


jcjimenez (approved):
LGTM

src/resolvers-cassandra/Settings/queries.js
```diff
@@ -47,37 +48,31 @@ function terms(args, res) { // eslint-disable-line no-unused-vars
...
-  return new Promise((resolve, reject) => {
-    const siteId = args.siteId;
-    if (!siteId) {
-      return reject('No site id to fetch specified');
-    }
-    
-    const siteById = 'SELECT * FROM fortis.sitesettings WHERE sitename = ?';
-    cassandraConnector.executeQuery(siteById, [siteId])
+  return new Promise((resolve, reject) => {    
+    const siteByIdQuery = 'SELECT * FROM fortis.sitesettings';
```
Smarker, line 68: Since we don't use `siteId` anymore, this could just be called `siteQuery`

src/resolvers-cassandra/Settings/index.js
```diff
@@ -20,7 +20,7 @@ module.exports = {
   removeKeywords: mutations.removeKeywords,
   addKeywords: mutations.addKeywords,
 
-  terms: queries.terms,
+  siteTerms: queries.siteTerms,
```
Smarker, line 23: This might be more clear if it were named `siteTopics` so that it is consistent with `cassandra-setup.cql`.

Smarker (approved):
LGTM

PR #132, opened 2017-09-11
Bug Bash Day II

erikschlegel:


src/resolvers-cassandra/Messages/queries.js
```diff
@@ -55,12 +56,15 @@ function queryEventsTable(eventIdResponse, args) {
 
     if(eventIdResponse.rows.length){
       cassandraConnector.executeQuery(eventsQuery, eventsParams)
-      .then(rows => resolve({
+      .then(rows => {
+        let test = 1;
```
Smarker: I think you forgot to remove the variable `test` here.

src/resolvers-cassandra/shared.js
```diff
@@ -140,6 +174,7 @@ module.exports = {
   tilesForBbox,
   tilesForLocations,
   limitForInClause,
+  getsiteDefintion,
```
Smarker, line 177: Could change this to `getSiteDefinition`

Smarker (approved):
LGTM

PR #137, opened 2017-09-12
Inline CSV report creation

c-w:
Due to how the frontend project is set up, this approach will be easier to consume from the frontend than making a separate graphql call since we can just toggle the csv include via flux.


jcjimenez (approved):
LGTM

PR #141, opened 2017-09-13
Send service bus message on edit or create of site.

Smarker:


src/clients/servicebus/ServiceBusClient.js
```diff
@@ -5,21 +5,25 @@ const azure = require('azure-sb');
...
  * @param {string} queue
  * @param {string} message
  * @returns {Promise}
  */
 function sendStringMessage(queue, message) {
   return new Promise((resolve, reject) => {
+    let client;
+    try {
+      client = azure.createServiceBusService(SERVICE_BUS_CONNECTION_STRING);
+      if (!client) return reject('Failed to create service bus service. No service bus connection string provided.');
```
c-w: Nit: This can live outside of the try block so that you don't risk catching spurious exceptions.

src/clients/servicebus/ServiceBusClient.js
```diff
@@ -5,21 +5,25 @@ const azure = require('azure-sb');
...
-let client = azure.createServiceBusService(SERVICE_BUS_CONNECTION_STRING);
-
 /**
  * @param {string} queue
  * @param {string} message
  * @returns {Promise}
  */
 function sendStringMessage(queue, message) {
   return new Promise((resolve, reject) => {
+    let client;
```
c-w: Nit: You can do the cheap checks below (message type, message length, etc.) first before trying to create the client which is a more expensive call.

src/resolvers-cassandra/Settings/mutations.js
```diff
@@ -86,6 +89,9 @@ function editSite(args, res) { // eslint-disable-line no-unused-vars
         ]
       }]);
     })
+    .then(() => {
+      serviceBusClient.sendMessages(SERVICE_BUS_CONFIG_QUEUE, JSON.stringify({'dirty': 'settings'}));
```
c-w: This is really just sending a single message so maybe it should be renamed to `sendMessage`?

c-w (approved):

src/resolvers-cassandra/Settings/mutations.js
Smarker: I agree, this is part of `azure-sb` though, not my `ServiceBusClient`.

src/clients/servicebus/ServiceBusClient.js
Smarker: yes, good point

PR #151, opened 2017-09-16
Prevent missing sb connection string from crashing the site.

Smarker:
If the connection string is missing or malformed, the services project will crash without this fix. 

**Will need an alert on the frontend if the service bus connection string is missing or malformed.**

### DIFF
--- a/src/clients/servicebus/ServiceBusClient.js
+++ b/src/clients/servicebus/ServiceBusClient.js
@@ -5,8 +5,6 @@ const azure = require('azure-sb');
 const trackDependency = require('../appinsights/AppInsightsClient').trackDependency;
 const SERVICE_BUS_CONNECTION_STRING = process.env.FORTIS_SB_CONN_STR;
 
-const client = azure.createServiceBusService(SERVICE_BUS_CONNECTION_STRING);
-
 /**
  * @param {string} queue
  * @param {string} message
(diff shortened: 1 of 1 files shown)


src/clients/servicebus/ServiceBusClient.js
```diff
@@ -19,6 +17,13 @@ function sendStringMessage(queue, message) {
       return reject('No message to be sent to service bus.');
     }
 
+    let client;
```
erikschlegel, line 20: Please raise an app insight error event in addition to rejecting the `Promise` https://github.com/Microsoft/ApplicationInsights-node.js#track-custom-telemetry

erikschlegel (changes requested):
we should be raising exceptions to application insights as discussed on fridays call. Please re-submit this PR with exception dependency tracking code in place. Also please provide a screenshot from the application insights dashboard showing this exception being raised with the severity level and useful properties indicating the origin of the exception. You should add a field to the exception payload(ie `trackException`) indicating the level of severity, so we can set up the appropriate monitors. https://github.com/Microsoft/ApplicationInsights-node.js/blob/0131e35d29c268c8c99368a7a3a6435828186569/Tests/Library/Client.tests.ts#L124

erikschlegel (approved):

PR #152, opened 2017-09-17
Add modify and remove trusted sources services.

Smarker:
These services are needed for the admin `trusted sources` tab.

image

image

**Corresponding Interfaces PR:** https://github.com/CatalystCode/project-fortis-interfaces/pull/76

### DIFF
--- a/src/resolvers-cassandra/Settings/index.js
+++ b/src/resolvers-cassandra/Settings/index.js
@@ -19,6 +19,8 @@ module.exports = {
   removeBlacklist: mutations.removeBlacklist,
   removeKeywords: mutations.removeKeywords,
   addKeywords: mutations.addKeywords,
+  modifyTrustedSources: mutations.modifyTrustedSources,
+  removeTrustedSources: mutations.removeTrustedSources,
 
   siteTerms: queries.siteTerms,
   sites: queries.sites,
@@ -27,5 +29,6 @@ module.exports = {
   trustedTwitterAccounts: queries.trustedTwitterAccounts,
   facebookPages: queries.facebookPages,
   facebookAnalytics: queries.facebookAnalytics,
-  termBlacklist: queries.termBlacklist
+  termBlacklist: queries.termBlacklist,
+  trustedSources: queries.trustedSources
 };
--- a/src/resolvers-cassandra/Settings/mutations.js
+++ b/src/resolvers-cassandra/Settings/mutations.js
@@ -510,7 +510,7 @@ function removeBlacklist(args, res) { // eslint-disable-line no-unused-vars
 
     const termIds = termFilters.map(termFilter => termFilter.id);
     
-    const query = `
+    const mutation = `
     DELETE
     FROM fortis.blacklist
     WHERE id IN ?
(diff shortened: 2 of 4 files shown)


erikschlegel:
The mutation endpoint looks good...  Please merge the recently published PR which covers the `trustedsources` GQL endpoint that's used on the dashboard for the source autosuggest search control.

erikschlegel (changes requested):
The mutation endpoint looks good... Please merge the recently published PR which covers the trustedsources GQL endpoint that's used on the dashboard for the source autosuggest search control.

Smarker:
Yep! Just saw your PR, so I'll use that instead for the query, but I do think we still need these so I can add/delete rows for the trusted sources tab.

src/resolvers-cassandra/Settings/index.js
```diff
@@ -27,5 +29,6 @@ module.exports = {
   trustedTwitterAccounts: queries.trustedTwitterAccounts,
   facebookPages: queries.facebookPages,
```
erikschlegel, line 30: `facebookPages` is a list of trusted sources so not sure why this is still needed.

src/resolvers-cassandra/Settings/mutations.js
```diff
@@ -510,7 +510,7 @@ function removeBlacklist(args, res) { // eslint-disable-line no-unused-vars
 
     const termIds = termFilters.map(termFilter => termFilter.id);
```
erikschlegel, line 512: I couldn't help notice that we're performing a combination of update and insert statements for cassandra records which I dont fully understand. when you invoke an insert statement in cassandra existing records are automatically updated.

src/resolvers-cassandra/Settings/queries.js
```diff
@@ -187,6 +187,32 @@ function termBlacklist(args, res) { // eslint-disable-line no-unused-vars
...
+  return {
+    rowKey: getTrustedSourcesRowKey(row.pipelinekey, row.externalsourceid, row.sourcetype, row.rank),
+    pipelineKey: row.pipelinekey,
+    externalSourceId: row.externalsourceid,
+    sourceType: row.sourcetype,
+    rank: row.rank
+  };
+}
+
+function trustedSources(args, res) { // eslint-disable-line no-unused-vars
```
erikschlegel, line 204: you can rebase from master to get the latest approach for querying `fortis.trustedsources`, but we should be filtering off `pipelinekeys` and avoiding an whole table query.

erikschlegel (changes requested):

src/resolvers-cassandra/Settings/index.js
Smarker: Just to confirm, `facebookPages` is not needed? I understand that now we have a `trustedSources` query so should that mean that `facebookPages` and `trustedTwitterAccounts` should be removed, since they are covered by the `trustedSources` query?

src/resolvers-cassandra/Settings/mutations.js
Smarker: When I do: 
`INSERT INTO trustedsources (pipelinekey,externalsourceid,sourcetype,rank,insertiontime) VALUES ('Twitter','MonteriaRadio38','user',100,1505313678830);` 
and then 
`INSERT INTO trustedsources (pipelinekey,externalsourceid,sourcetype,rank,insertiontime) VALUES ('twitter','MonteriaRadio38','user',100,1505313678830);` 
I get 2 records.

Smarker:
Will close this PR and provide a separate PR with Erik's most recent trusted sources query and schema changes.

**Updated PR (based off this PR#152 but with Erik's latest changes):** https://github.com/CatalystCode/project-fortis-services/pull/154

PR #153, opened 2017-09-17
Adding trusted sources service

erikschlegel:


src/resolvers-cassandra/Settings/index.js
```diff
@@ -22,6 +22,7 @@ module.exports = {
 
   siteTerms: queries.siteTerms,
   sites: queries.sites,
+  trustedsources: queries.trustedsources,
```
Smarker: No camel case?

Smarker (approved):
LGTM - I'll update the admin trusted sources tab so it uses this service instead of the one I had.

src/resolvers-cassandra/Settings/index.js
erikschlegel: good point. I'll change it to Camel Case for consistency.

PR #154, opened 2017-09-17
Modify and remove trusted sources.

Smarker:
Also rearranged stream helper functions so that they would be closer to the streams query.

Will remove unnecessary `facebookPages` query (and possibly other unused queries/mutations) in a separate PR.


Smarker:
**removeTrustedSources:**
image

**modifyTrustedSources:**
image

erikschlegel (approved):

PR #158, opened 2017-09-28
Adding temporary workaround for blacklist filter

erikschlegel:
This will be removed once @NathanielRose has the blacklist filter fully integrated in Fortis.


jcjimenez (approved):
LGTM

PR #162, opened 2017-10-12
Streams final

Smarker:
For issue https://github.com/CatalystCode/project-fortis-interfaces/issues/94
App insights logging:
Modify Streams:
image

Query Streams:
image

No need for remove streams service, since the user should only be able to edit stream param values and enable or disable a stream.


jcjimenez (approved):
LGTM

PR #166, opened 2017-10-31
Fix create_site.py failing

c-w:
The createSite call was failing with an error of "Unmatched column names/values". This is due to a regression introduced in 872e1ae8 which added a param to the fortis.watchlist insertion cassandra query without adding a placeholder in the query CQL.

Additionally, the topic category is required to always be non-null so we default it to an empty string if the category is not set in the blob watchlist terms.

Tagging @erikschlegel since he requested the fix.

Here is the createSite mutation succeeding after the fix:

image

Here is the create_site.py script succeeding after the fix:

image


Smarker (approved):
LGTM

PR #167, opened 2017-11-28
Dockerize

### LINKED ISSUE #164: GQL Service Documentation
We need to give some much needed TLC to this repos README . The readme should provide clear guidance for new Fortis developers.

Documentation should cover the following criteria

 1. Installing the fortis service project and it's dependencies, then starting the node server. 
 2. Documenting all Queries and Mutations within the `Messages`, `Settings`, `Tiles` and `Edges` GQL endpoints. 
 3. The documentation should include a brief description for each query and mutation, including its arguments and return type. 
 4. We should also include a sample GQL query and mutation call for each service call.

c-w:
See https://github.com/CatalystCode/project-fortis-pipeline/issues/228

This also resolves https://github.com/CatalystCode/project-fortis-services/issues/164 since the readme update points users directly to the GraphiQL test consoles; from there, everything in GraphQL should be self-documenting.

### DIFF
--- a/docker/project-fortis-services/run-server.sh
+++ b/docker/project-fortis-services/run-server.sh
@@ -1,0 +1,33 @@
+#!/usr/bin/env sh
+
+cassandra_exec() {
+  /opt/cassandra/bin/cqlsh \
+    --request-timeout=3600 \
+    --username="$CASSANDRA_USERNAME" \
+    --password="$CASSANDRA_PASSWORD" \
+    "$CASSANDRA_CONTACT_POINTS"
+}
+
+# wait for cassandra to start
+while ! cassandra_exec; do
+  echo "Cassandra not yet available, waiting..."
+  sleep 10s
+done
+echo "...done, Cassandra is now available"
+
+# set up cassandra schema
+if [ -n "$FORTIS_CASSANDRA_SCHEMA_URL" ]; then
+  echo "Got Fortis schema definition at $FORTIS_CASSANDRA_SCHEMA_URL, ingesting..."
+  wget -qO- "$FORTIS_CASSANDRA_SCHEMA_URL" | cassandra_exec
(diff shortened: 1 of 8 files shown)


docker/project-fortis-services/run-server.sh
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env sh
```
jcjimenez, line 1: Any chance you can add a `set -e` so that it exists on first failure?

```
#!/usr/bin/env sh
set -e
```

docker/project-fortis-services/run-server.sh
```diff
@@ -0,0 +1,33 @@
...
+while ! cassandra_exec; do
+  echo "Cassandra not yet available, waiting..."
+  sleep 10s
+done
+echo "...done, Cassandra is now available"
+
+# set up cassandra schema
+if [ -n "$FORTIS_CASSANDRA_SCHEMA_URL" ]; then
+  echo "Got Fortis schema definition at $FORTIS_CASSANDRA_SCHEMA_URL, ingesting..."
+  wget -qO- "$FORTIS_CASSANDRA_SCHEMA_URL" | cassandra_exec
```
jcjimenez, line 21: Excellent! At some point, I think we could move these DDL statements into database migrations running on this project (fortis-services).

jcjimenez (approved):
LGTM!

docker/project-fortis-services/run-server.sh
c-w: Great comment. I'm usually the first one to advocate for Bash strict mode (`set -euo pipefail`). However, some of the commands below are expected to fail (e.g. when waiting for Cassandra to come up) so I think it's cleaner to not enable strict-mode here since otherwise we'd have to sprinkle a bunch of `|| true` across the code which may confuse folks with less Bash experience.

